PR #7, opened 2025-07-18
add version verification in initccm

thweetkomputer:
- [ ] Document changes

### DIFF
--- a/include/cc/cc_req_base.h
+++ b/include/cc/cc_req_base.h
@@ -87,6 +87,11 @@ public:
         assert(false && "Unimplemented virtual method");
     }
 
+    virtual uint64_t SchemaVersion() const
+    {
+        return 0;
+    }
+
 protected:
     CcRequestBase() = default;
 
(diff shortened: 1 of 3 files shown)


include/cc/cc_req_base.h
```diff
@@ -87,6 +87,11 @@ struct CcRequestBase
         assert(false && "Unimplemented virtual method");
     }
 
+    virtual uint64_t SchemaVersion() const
```
MrGuin, line 90: There is no point making this virtual. Why not declare schema_version_ in CcRequestBase and make this uint64_t SchemaVersion()`.

src/cc/cc_shard.cpp
```diff
@@ -1662,6 +1662,21 @@ const CatalogEntry *CcShard::InitCcm(const TableName &table_name,
...
+        {
+            // For DDL operations (e.g., `flushdb` in Redis protocol), if one tx
+            // processor completes the operation earlier, it could potentially
+            // read data from other tx processors by simply acquiring a read
+            // lock on itself. However, other processors might not have
+            // completed the DDL operation yet, meaning their schema version
+            // could still be old. Initiating ccm with the old version leads to
+            // conflicts. Therefore, we explicitly verify the schema version
+            // here to prevent such cases.
+            requester->AbortCcRequest(CcErrorCode::FORCE_FAIL);
```
MrGuin, line 1677: Use error code REQUESTED_TABLE_SCHEMA_MISMATCH.

thweetkomputer: fixed

PR #13, opened 2025-07-24
make non-shared-storage clean ccm and upsert kv idempotent

### LINKED ISSUE #12: [Bug]: Non-shared-storage flushdb recovery bug.
When a DDL txn recovers from commit log, we have to ensure that the memory map clean and kv clean be  idempotent because before failure, these operatons could have been executed once and the partipants could have released the write lock and received new data. The new data of new table cannot be deleted by the recovered DDL txn. In the case of non-shared-storage EloqKV, this condition is not properly enforced. While CleanCcm validates the schema version and prevents repeated ccmap cleanup, the upsert logic in the key-value storage does not consistently follow the same version checks, potentially leading to crash and inconsistencies.

MrGuin:
Fix #12.

### DIFF
--- a/include/cc/cc_request.h
+++ b/include/cc/cc_request.h
@@ -3356,8 +3356,7 @@ public:
             slices_to_scan_.reserve(old_slices_delta_size->size());
             std::for_each(old_slices_delta_size->begin(),
                           old_slices_delta_size->end(),
-                          [&](decltype(*old_slices_delta_size->begin()) &elem)
-                          {
+                          [&](decltype(*old_slices_delta_size->begin()) &elem) {
                               slices_to_scan_.emplace_back(
                                   std::move(elem.first.GetShallowCopy()));
                           });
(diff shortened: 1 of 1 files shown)


include/cc/cc_request.h
```diff
@@ -5130,6 +5130,23 @@ struct KickoutCcEntryCc : public TemplatedCcRequest<KickoutCcEntryCc, Void>
...
+                const CatalogEntry *catalog_entry =
+                    ccs.GetCatalog(*table_name_, node_group_id_);
+                if (catalog_entry == nullptr)
+                {
+                    //  Fetch catalog
+                    ccs.FetchCatalog(
+                        *table_name_, node_group_id_, ng_term, this);
+                    return false;
+                }
+                if (catalog_entry->schema_version_ >= clean_ts_)
```
thweetkomputer, line 5142: what if `catalog_entry->schema_version_  < clean_ts_`

thweetkomputer (approved):

include/cc/cc_request.h
```diff
@@ -5130,6 +5130,23 @@ struct KickoutCcEntryCc : public TemplatedCcRequest<KickoutCcEntryCc, Void>
...
+                {
+                    // This is an out-dated request. The table has already been
+                    // cleaned and updated.
+                    return SetFinish();
+                }
```
JoJossd, line 5147: should CleanCcMap here if recovering from commit_log?

PR #16, opened 2025-07-25
Ckpt refactor

liunyl:
Buffer data from multiple data sync tasks in a flush buffer and flush them all at once.
We do this for the following reasons.
1. For rocksdb, write in a bigger batch then call PersistKv generates less but bigger sst files, which results in smaller compaction pressure on level 0.
2. For eloqstore, write in a bigger batch for each partition improves overall write efficiency.

### DIFF
--- a/include/store/data_store_handler.h
+++ b/include/store/data_store_handler.h
@@ -99,12 +99,11 @@ public:
 
     virtual void ScheduleTimerTasks() {};
     /**
-     * @brief flush entries in \@param batch to base table or skindex table in
-     * data store, stop and return false if node_group is not longer leader.
-     * @param batch
-     * @param table_name base table name or sk index name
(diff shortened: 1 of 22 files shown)


include/store/data_store_handler.h
```diff
@@ -106,11 +107,10 @@ class DataStoreHandler
      * @param node_group
      * @return whether all entries are written to data store successfully
      */
-    virtual bool PutAll(std::vector<txservice::FlushRecord> &batch,
-                        const txservice::TableName &table_name,
-                        const txservice::TableSchema *table_schema,
-                        uint32_t node_group) = 0;
-
+    virtual bool PutAll(std::unordered_map<std::string_view,
```
liunyl, line 110: update comment

PR #50, opened 2025-08-04
remove range part macro from ckpt related code

liunyl:
- [ ] Document changes


include/cc/local_cc_shards.h
```diff
@@ -2211,33 +2192,18 @@ class LocalCcShards
...
             : node_group_id_(node_group_id),
               node_group_term_(node_group_term),
               table_name_(table_name, table_type, table_engine),
-              range_id_(range_id)
+              id_(id)
         {
         }
-#else
-        explicit TaskLimiterKey(NodeGroupId node_group_id,
-                                int64_t node_group_term,
```
lokax, line 2229: We should distinguish task limiters for RANGE and HASH. otherwise, when the range id equals the core id, the hash/range tasks will be skipped.

PR #59, opened 2025-08-07
Pass range_id into ReadCc when do read

lzxddz:
- [ ] Document changes


include/cc/template_cc_map.h
```diff
@@ -1534,6 +1534,7 @@ class TemplateCcMap : public CcMap
                                 RecordSchema(),
                                 schema_ts_,
                                 table_schema_->GetKVCatalogInfo(),
```
liunyl, line 1536: FetchRecord which is used to fetch archive should also use this range id

liunyl (approved):

PR #92, opened 2025-08-22
Refactor scan open and scan next operation

lokax:
- [ ] Document changes

### DIFF
--- a/src/cc/local_cc_handler.cpp
+++ b/src/cc/local_cc_handler.cpp
@@ -1188,62 +1188,8 @@ void txservice::LocalCcHandler::ScanOpenLocal(
     scanner_ptr->lock_type_ = LockTypeUtil::DeduceLockType(
         scanner_ptr->cc_op_, iso_level, proto, false);
 
-    // TODO(lokax):
     hd_res.SetFinished();
     return;
-
-    /*
-    ScanCache *shard_scan_cache = scanner_ptr->AddShard(shard_code);
-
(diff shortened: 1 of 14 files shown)


src/cc/local_cc_handler.cpp
```diff
@@ -1196,59 +1208,13 @@ void txservice::LocalCcHandler::ScanOpenLocal(
...
-    }
-    else
-    {
-#ifdef EXT_TX_PROC_ENABLED
-        hd_res.SetToBlock();
-#endif
-        local_shard.Enqueue(scan_open_cc_req);
-    }
+    hd_res.SetFinished();
+    return;
```
liunyl, line 1212: we should allow hash partition scan to specify start key/end key. It will be passed into scan of each bucket

src/tx_execution.cpp
```diff
@@ -2496,10 +2525,72 @@ void TransactionExecution::PostProcess(ScanOpenOperation &scan_open)
...
+            // <NodeGroupId, <core_idx, vector<bucket_id>>
+            // group by node group and core idx
+            absl::flat_hash_map<
+                NodeGroupId,
+                absl::flat_hash_map<uint16_t, std::vector<uint16_t>>>
+                grouped;
+            for (size_t bucket_id = 0; bucket_id < total_range_buckets;
+                 ++bucket_id)
+            {
+                NodeGroupId bucket_owner =
```
liunyl, line 2543: Check if leader term is valid before accessing structs in local cc shards

PR #108, opened 2025-08-29
Create snapshot branch

githubzilla:
- [ ] Document changes


include/proto/cc_request.proto
```diff
@@ -1431,5 +1439,9 @@ message CcMessage {
         KeyObjectStandbyForwardRequest key_obj_standby_forward_req=64;
         InvalidateTableCacheRequest invalidate_table_cache_req = 65;
         InvalidateTableCacheResponse invalidate_table_cache_resp = 66;
+        CreateSnapshotRequest create_snapshot_request = 67;
```
liunyl, line 1442: Remove these

include/store/snapshot_manager.h
```diff
@@ -93,8 +99,10 @@ class SnapshotManager
 
     const std::string backup_path_;
     std::mutex backup_tasks_mux_;
-    static const size_t MaxBackupTaskCount = 100;
+    // Queue of backup tasks whose status can be queried
     std::deque<txservice::remote::CreateBackupRequest *> backup_task_queue_;
+    // Limit the max number of backup tasks in queue
+    static const size_t MaxBackupTaskCount = 100;
```
liunyl, line 105: There should only be 1 backup/snapshot request at a time

include/store/snapshot_manager.h
```diff
@@ -93,8 +99,10 @@ class SnapshotManager
 
     const std::string backup_path_;
     std::mutex backup_tasks_mux_;
-    static const size_t MaxBackupTaskCount = 100;
+    // Queue of backup tasks whose status can be queried
```
liunyl, line 102: who is the owner of these reqs?

src/checkpointer.cpp
```diff
@@ -435,24 +439,33 @@ void Checkpointer::Run()
...
             break;
         }
 
         last_checkpoint_ts_ = std::chrono::high_resolution_clock::now();
+        is_doing_ckpt_ = true;
         lk.unlock();
         Ckpt(false);
         lk.lock();
+        // notify all waiting that one round checkpoint is done.
+        is_doing_ckpt_ = false;
```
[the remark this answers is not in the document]
liunyl: we also need to consider DDL. some of the DDL like create index might flush data by itself. We need to reject snapshot/backup request when there're concurrent DDL

include/store/snapshot_manager.h
githubzilla: Yes, there is only 1 thread of the backup_worker_ thread pool, which guarantee only 1 backup request at a time.

src/checkpointer.cpp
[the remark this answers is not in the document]
githubzilla: As discussed, we will implement the snapshot assuming there are no ongoing CKPT, DDL, or Bucket Migration processes. Meanwhile, we will work on defining the details of the new approach that combines data store snapshots with replay logs.

PR #134, opened 2025-09-16
Update ng leader cache from multi-stage log response.

MrGuin:
So that the coordinator knows the latest leader info when some participant fails over.


MrGuin:
Is it enough to just compare the ip stored in log service with the host name stored in ng_configs_? Would it happen that two nodes in the same group share the same hostname? @liunyl

thweetkomputer (approved):

liunyl:
@coderabbitai review

PR #139, opened 2025-09-19
 Improve hash-partition data sync scheduling and memory control.

thweetkomputer:
- Add per-worker coordination (data_sync_done_, scan_concurrency_) so only a bounded number of hash-partition scans run concurrently, avoiding contention.
  - Split ScanSliceDeltaSizeCc into range/hash variants, letting the hash version track key deltas and per-core memory usage for better sizing.
  - Pull hash partition IDs from hash & 0x3FF, reserve sync buffers, and partition scans according to UpdatedMemory() and flush-buffer capacity to prevent over-allocation.
  - In-place construct FlushRecord payloads and move-owned callbacks/status in DataSyncTask to reduce copies.
  - Requeue failed hash sync tasks instead of dropping them, keeping checkpoint retries reliable.

## Summary by CodeRabbit

- New Features
  - Partition-aware, multi-core data synchronization and a new flush-buffer size report.
- Performance
  - Parallel delta-size scans, improved memory accounting, and optimized key-to-partition mapping.
- Bug Fixes
  - More accurate partition selection and safer lambda captures for stable callbacks.
- Refactor
  - Split range vs. hash partition scan flows and added async coordination for hash-partition scans.
- Style
  - Minor comment typo corrected.

### DIFF
--- a/include/cc/cc_map.h
+++ b/include/cc/cc_map.h
@@ -80,7 +80,8 @@ struct EscalateStandbyCcmCc;
 struct RestoreCcMapCc;
 struct InvalidateTableCacheCc;
 struct SampleSubRangeKeysCc;
-struct ScanSliceDeltaSizeCc;
+struct ScanSliceDeltaSizeCcForRangePartition;
+struct ScanDeltaSizeCcForHashPartition;
 
 enum struct ScanType : uint8_t
 {
@@ -203,7 +204,8 @@ public:
     virtual bool Execute(RestoreCcMapCc &req) = 0;
     virtual bool Execute(InvalidateTableCacheCc &req) = 0;
     virtual bool Execute(SampleSubRangeKeysCc &req) = 0;
-    virtual bool Execute(ScanSliceDeltaSizeCc &req) = 0;
+    virtual bool Execute(ScanSliceDeltaSizeCcForRangePartition &req) = 0;
+    virtual bool Execute(ScanDeltaSizeCcForHashPartition &req) = 0;
 
     virtual size_t size() const = 0;
     virtual size_t NormalObjectSize()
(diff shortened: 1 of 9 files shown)


include/cc/cc_map.h
```diff
@@ -80,7 +80,8 @@ struct EscalateStandbyCcmCc;
 struct RestoreCcMapCc;
 struct InvalidateTableCacheCc;
 struct SampleSubRangeKeysCc;
-struct ScanSliceDeltaSizeCc;
+struct ScanSliceDeltaSizeCcForRangePartition;
+struct ScanSliceDeltaSizeCcForHashPartition;
```
liunyl, line 84: slice is only used in range partition. For hash partition just name it ScanDeltaSizeCcForHashPartition

include/data_sync_task.h
```diff
@@ -303,6 +304,12 @@ struct FlushDataTask
         return pending_flush_size_;
     }
 
+    size_t GetFlushBufferSize()
+    {
+        std::lock_guard<bthread::Mutex> lk(flush_task_entries_mux_);
```
liunyl, line 309: no mutex is needed here since this value is never updated

src/cc/local_cc_shards.cpp
```diff
@@ -2813,36 +2817,92 @@ void LocalCcShards::DataSyncWorker(size_t worker_idx)
...
+                               scan_concurrency_.load(
+                                   std::memory_order_relaxed) ||
+                           data_sync_worker_ctx_.status_ !=
+                               WorkerStatus::Active;
+                });
+            task_worker_lk.unlock();
             DataSyncForHashPartition(std::move(task), worker_idx);
+            task_worker_lk.lock();
+
+            data_sync_done_[worker_idx] = true;
```
liunyl, line 2883: does this starve range partition data sync if running together?

PR #189, opened 2025-11-05
Update cluster backup rpc interface

liunyl:
- [ ] Document changes

## Summary by CodeRabbit

* **Refactor**
  * CreateBackup and related cluster backup endpoints now return richer per-node responses that include status, backup files and timestamps instead of simple status-only payloads.
  * Cluster backup flows and utilities updated to aggregate and propagate these detailed per-node responses so backup creation and status queries return consolidated backup details.


githubzilla (approved):
+1

PR #192, opened 2025-11-06
store writer txn instead of write_req when write_req is processed

MrGuin:
## Summary by CodeRabbit

* **Refactor**
  * Optimized internal writer state handling in the concurrency control layer for improved memory efficiency and transaction tracking; no public API changes.
* **Bug Fixes**
  * Made catalog commit handling more robust by explicitly cleaning updated entries and updating schema state during truncation.
  * Replaced an assertion with a logged warning and graceful READ_CATALOG_CONFLICT return when schema write-lock assumptions are not met.


githubzilla (approved):
+1

PR #203, opened 2025-11-11
Fix eloqkv report oom

yi-xmu:
1. Normally, when a `ccrequest` can not find enough memory when `FindEmplace`, it triggers a shard clean operation and waits until free memory becomes available. However, if the transaction already holds a range read lock, the `ccrequest` should be aborted with error code `OUT_OF_MEMORY` if no free memory is available; otherwise, it may block range splitting.
2. Remove the macro `ON_KEY_OBJECT`.

## Summary by CodeRabbit

* **New Features**
  * Per-request "abort-on-OOM" flag added so individual operations can opt into early abort behavior.
  * Transaction execution exposes range-read-lock state to influence abort-on-OOM decisions.

* **Bug Fixes**
  * OOM handling made deterministic: requests now either re-enqueue or abort immediately based on the new flag.

* **Refactor**
  * Memory wait-list and dequeue flow simplified; new path to abort waiting requests when memory frees.

### DIFF
--- a/include/cc/cc_request.h
+++ b/include/cc/cc_request.h
@@ -377,7 +377,8 @@ public:
                CcHandlerResult<std::vector<AcquireKeyResult>> *res,
                uint32_t hd_res_idx,
                CcProtocol proto,
-               IsolationLevel iso_level)
+               IsolationLevel iso_level,
+               bool abort_if_oom)
     {
         uint32_t ng_id = Sharder::Instance().ShardToCcNodeGroup(key_shard_code);
         TemplatedCcRequest<AcquireCc, std::vector<AcquireKeyResult>>::Reset(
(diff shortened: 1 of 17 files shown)


include/cc/cc_request.h
```diff
@@ -7321,6 +7380,12 @@ struct RequestAborterCc : public CcRequestBase
     {
         for (CcRequestBase *req : reqs_)
         {
+            if (err_code_ == CcErrorCode::OUT_OF_MEMORY && !req->AbortIfOom())
```
liunyl: when do we use the request aborter cc on OOM error? This cc req is usually used when we want to abort deadlock cc req

yi-xmu: If `LoadSlice` fails (`FillStoreSlice` failed), call the waiting request abort.

PR #240, opened 2025-11-25
fix command forwarding to standby node

MrGuin:
* set_has_overwrite if the cce is in Deleted status when forwarding the command to standby node;
* fix the bug that datasyncscan is wrongly marked success when cces HasBufferCommandList and are skipped.

Related PR:
https://github.com/eloqdata/eloqkv/pull/305

## Summary by CodeRabbit

* **Bug Fixes**
  * Deleted objects are now flagged so standby propagation treats them as overwrites.
  * Checkpoint success now requires no skipped entries; standby sync retry timeout increased for more reliable synchronization.

* **Enhancements**
  * Data-sync now continues when buffered commands exist, improving replay coverage and reducing missed updates.
  * New status tracking for skipped entries prevents unintended log truncation and surfaces warnings.

### DIFF
--- a/include/data_sync_task.h
+++ b/include/data_sync_task.h
@@ -62,7 +62,7 @@ struct DataSyncStatus
     void SetEntriesSkippedAndNoTruncateLog()
     {
         std::lock_guard<std::mutex> lk(mux_);
-        has_skipped_entries = true;
+        has_skipped_entries_ = true;
         need_truncate_log_ = false;
     }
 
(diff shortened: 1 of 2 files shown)


include/cc/template_cc_map.h
```diff
@@ -5890,15 +5890,14 @@ class TemplateCcMap : public CcMap
...
-                            continue;
-                        }
+                        DLOG(INFO)
+                            << "cce has buffer command, skip "
+                               "ckpt cce: "
+                            << cce->KeyString()
+                            << ", cce status: " << int(cce->PayloadStatus())
+                            << ", BufferCommand: " << cce->BufferedCommandList()
+                            << ", data_sync_ts_: " << req.data_sync_ts_;
+                        replay_cmds_notnull = true;
```
liunyl: If we're sending this snapshot to standby for sync, we need to wait for all buffered cmds to be cleared. Otherwise the unapplied cmds will be missed on standby node

liunyl (approved):

PR #283, opened 2025-12-10
Range read block on rw conflict

githubzilla:
- [ ] Document changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Range read requests now block and wait for locks (instead of aborting), preserving read locks and queueing retries; remote requesters receive acknowledgments when applicable.
  * Error paths now enter a safe waiting state to allow deadlock detection to observe real wait conditions.

* **Chores**
  * Added per-second throttling for global deadlock checks and made deadlock-check signaling thread-safe.
  * Improved diagnostic logging and fault-injection hooks for transaction flows.

### DIFF
--- a/tx_service/include/dead_lock_check.h
+++ b/tx_service/include/dead_lock_check.h
@@ -242,11 +242,5 @@ protected:
     uint32_t check_node_id_;
     // If the dead lock check is requested by this node.
     std::atomic<bool> requested_check_{false};
-    // Last throttled check request time (microseconds)
-    // Used to throttle deadlock check requests to once per second globally for
-    // all transactions
-    uint64_t last_throttled_check_time_{0};
-    static constexpr uint64_t CHECK_THROTTLE_INTERVAL_ =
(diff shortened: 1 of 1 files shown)


tx_service/include/cc/range_cc_map.h
```diff
@@ -324,19 +325,27 @@ class RangeCcMap : public TemplateCcMap<KeyT, RangeRecord, true, false>
...
+
+            // Block instead of abort - similar to regular data reads
+            // Keep the bucket read lock (don't release it) - this matches the
+            // pattern of regular data reads which don't release locks when
+            // blocking. The bucket lock is a read lock, so it doesn't prevent
+            // other readers. When the request resumes, we already have the
+            // bucket lock and can immediately retry the range lock acquisition.
+            // If the read request comes from a remote node, send
+            // acknowledgement
+            if (!req.IsLocal())
```
liunyl: this should never happen. range cc map ReadCc should only be initiated by local node

PR #323, opened 2025-12-25
Index mem

liunyl:
- [ ] Document changes

## Summary by CodeRabbit

* **Chores**
  * Improved memory cleanup in batch operations and reduced retained capacity after clears.
  * Tightened memory/quota calculations for data synchronization by including additional in-memory footprints.
  * Broadened lock-handling during transaction recovery to treat more non-writable locks uniformly.
  * Increased the default load-factor used for range partitioning decisions.

### DIFF
--- a/tx_service/src/cc/local_cc_shards.cpp
+++ b/tx_service/src/cc/local_cc_shards.cpp
@@ -4095,13 +4095,18 @@ void LocalCcShards::DataSyncForRangePartition(
             }
 
             // The cost of FlushRecord also needs to be considered.
+            FlushRecord flush_record;
+            size_t flush_record_size = mi_malloc_usable_size(&flush_record);
+#ifdef WITH_JEMALLOC
+            flush_record_size = sizeof(FlushRecord);
+#endif
             for (size_t i = 0; i < cc_shards_.size(); ++i)
             {
-                flush_data_size +=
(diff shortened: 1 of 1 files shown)


xiexiaoy:
The calculation of memory usage depends on sizeof, which is incorrect. Better call `size_t malloc_usable_size(void *ptr);`.

tx_service/src/cc/local_cc_shards.cpp
```diff
@@ -4094,6 +4094,21 @@ void LocalCcShards::DataSyncForRangePartition(
                 flush_data_size += flush_data_size_per_core;
             }
 
+            // The cost of FlushRecord also needs to be considered.
+            FlushRecord flush_record;
+            size_t flush_record_size = mi_malloc_usable_size(&flush_record);
+#ifdef WITH_JEMALLOC
+            flush_record_size = sizeof(FlushRecord);
+#endif
```
xiexiaoy: Please take a look at the figure I pasted in eloqdata/project_tracker#73.

std::vector<FlushRecord> data_sync_vecs, archive_vecs and mv_base_vecs might allocs memory more than calculated value.

I suggest using mi_malloc_usable_size to get the memory usage of those vectors, not the variable flush_record.

xiexiaoy (approved):

PR #326, opened 2025-12-26
fix standby OnLeaderStart and FetchCatalog from primary

MrGuin:
* mark that this node is becoming leader and discard the standby msgs in OnLeaderStart before wait InflightStandbyReqCount;
* FetchCatalog from primary aborts the fetch_cc if rpc fails or node is becoming leader.

## Summary by CodeRabbit

* **New Features**
  * Public API to track when a standby node is becoming leader, with initialization support.

* **Bug Fixes**
  * Improved error logging and clearer aborts when leader is unreachable during catalog fetches.
  * Added validation gates to prevent actions during standby→leader transitions, including gating inflight forwarding and remote fetches.

* **Chores**
  * Introduced a new error code to surface leader-unreachable conditions.

### DIFF
--- a/tx_service/src/sharder.cpp
+++ b/tx_service/src/sharder.cpp
@@ -174,7 +174,8 @@ int Sharder::Init(
 
         standby_node_term_cache_.store(-1, std::memory_order_relaxed);
         candidate_standby_node_term_cache_.store(-1, std::memory_order_relaxed);
-        standby_becoming_leader_term_cache_.store(-1, std::memory_order_relaxed);
+        standby_becoming_leader_term_cache_.store(-1,
+                                                  std::memory_order_relaxed);
 
         if (ng_configs != nullptr)
         {


tx_service/include/sharder.h
```diff
@@ -681,6 +681,28 @@ class Sharder
...
     }
 
+    void SetStandbyBecomingLeaderNodeTerm(int64_t leader_term)
+    {
+        if (!cc_nodes_init_.load(std::memory_order_acquire))
+        {
+            return;
+        }
+
+        standby_becoming_leader_term_cache_.store(leader_term,
```
liunyl: seems like this only need to be a boolean

tx_service/include/rpc_closure.h
```diff
@@ -973,6 +975,9 @@ class FetchCatalogClosure : public ::google::protobuf::Closure
                 stub.FetchCatalog(&cntl_, &request_, &response_, this);
                 return;
             }
+            DLOG(ERROR) << "FetchCatalogClosure fails, abort fetch_cc_";
+            fetch_cc_->SetFinish(RecordStatus::Deleted,
```
liunyl: need to fix ForwardStandbyCc abort logic. ForwardStandbyCc cannot be discarded even if called abort cc req, it needs to be reenqueued into the cc req queue otherwise we might lost data.

liunyl (approved):

PR #332, opened 2026-01-02
Fix datasyncscan paused key with debug log

yi-xmu:
The paused key should be the end key of slice if it paused on the end of a slice.

## Bug Fixes
* Added bounds checking to prevent potential access violations when advancing through data slices during scan operations.
* Refined resumption logic for data scans across partitions to correctly position pause points at slice boundaries.


yi-xmu:
https://github.com/eloqdata/eloqsql/pull/198  
https://github.com/eloqdata/eloqdoc/pull/405

PR #406, opened 2026-02-10
Dirty memory ratio and size can trigger ckpt

githubzilla:
- [ ] Document changes

## Summary by CodeRabbit

* **New Features**
  * Proactive dirty-memory monitoring with configurable check interval and size threshold to auto-trigger checkpoints.
  * Periodic sampling of dirty keys and enhanced checkpoint logging for improved observability.

* **Bug Fixes / Improvements**
  * Replaced one-off memory-full notification with continuous dirty-memory tracking to reduce spurious checkpoint signals and improve stability.

### DIFF
--- a/tx_service/src/checkpointer.cpp
+++ b/tx_service/src/checkpointer.cpp
@@ -65,7 +65,12 @@ Checkpointer::Checkpointer(LocalCcShards &shards,
       checkpoint_interval_(checkpoint_interval),
       min_ckpt_request_interval_(min_ckpt_request_interval),
       last_checkpoint_ts_(std::chrono::system_clock::now()),
-      last_checkpoint_request_ts_(std::chrono::system_clock::now()),
+      // Initialize last_checkpoint_request_ts_ to a time point that is
+      // sufficiently in the past, so that the first checkpoint request can be
+      // processed immediately
+      last_checkpoint_request_ts_(
+          std::chrono::system_clock::now() -
+          std::chrono::seconds(2 * min_ckpt_request_interval)),
       ckpt_delay_time_(ckpt_delay_seconds * 1000000),
       ongoing_data_sync_cnt_(0),
       tx_service_(nullptr),


tx_service/include/checkpointer.h
```diff
@@ -152,10 +162,10 @@ class Checkpointer
     };
 
     LocalCcShards &local_shards_;
-    // protects request_ckpt_ and status_
+    // protects status_
     std::mutex ckpt_mux_;
     std::condition_variable ckpt_cv_;
-    bool request_ckpt_;
+    std::atomic<bool> request_ckpt_{false};
```
liunyl: why is this changed to atomic? atomic and mutex + cv does not work well together

githubzilla:
We have more work to do, since liuyl and I have new idea to control the frequency of the NotifyCKPT should be triggered
- Each ccshard will trigger CKPT individually, so we need to prevent the spam by check the request_ckpt_, but the mutex acquire too expensive, we change the type of request_ckpt_ as std::atomic, and using the release/acquire to guarantee the memory order.
- Before all DataSyncTask are finished on a core, the dirty memory footprint can be still high on that core, and the Ckpt will be triggered unnecessarily, we should avoid this issue.
   We keep track of the DataSyncTask for a core using a reference count, new Ckpt will be triggered only if the reference count is zero

tx_service/src/cc/cc_req_misc.cpp
```diff
@@ -1493,14 +1493,8 @@ bool ShardCleanCc::Execute(CcShard &ccs)
                 // waiting ccrequests.
                 ccs.AbortRequestsAfterMemoryFree();
 
-                // Notify the checkpointer thread to do checkpoint if there
-                // is not freeable entries to be kicked out from ccmap and
-                // if the shard is not doing defrag.
-                if (free_count_ == 0 && !shard_heap->IsDefragHeapCcOnFly() &&
```
liunyl: We should keep this as a safety check. This should not trigger. But if user set some terrible params like check interval = 1000000 and dirty ratio = 99%, we might be stuck at memory full and unable to trigger checkpoint

tx_service/include/cc/cc_shard.h
```diff
@@ -1291,6 +1299,14 @@ class CcShard
     size_t data_key_count_{0};
     // The number of committed dirty keys in data tables only.
     size_t dirty_data_key_count_{0};
+    // Counter for sampling dirty memory checks in AdjustDataKeyStats.
+    uint64_t adjust_stats_call_count_{0};
+
+    // Config for dirty memory checkpoint triggering.
+    uint64_t dirty_memory_check_interval_{1000};
+    uint64_t dirty_memory_size_threshold_mb_{0};
```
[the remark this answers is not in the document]
liunyl: I mean dirty_memory_threshold_bytes_ and dirty_memory_size_threshold_mb_

PR #484, opened 2026-06-01
feat: partition-level reopen for buffered commands on standby

thweetkomputer:
- Add delayed reopen queue in eloqstore with priority heap
- Add ReopenPartition virtual chain through DataStore/DataStoreService
- Add CcShard::RequestPartitionReopen with per-partition dedup
- Trigger reopen from FetchRecordCc, UploadTxCommandsCc, KeyObjectStandbyForwardCc, and ReplayLogCc when buffered commands remain after BackFill
- Fix BackFill condition (&& -> ||) to allow CCE update when KV has a newer version, needed when reopen loads fresh cloud data

Signed-off-by: Jerry Zhao

- [ ] Document changes

## Summary by CodeRabbit

* **New Features**
  * Read operations can request partition "reopen" before fetching, improving correctness during concurrent/standby workflows.
  * After committing buffered commands, the system may trigger an additional standby record fetch to ensure up-to-date data.

* **Chores**
  * New configurable auto-reopen pending timeout (default 10s) to tune reopen timing.


thweetkomputer:
@liangjchen Could you review this PR?

liangjchen:
> @liangjchen Could you review this PR?

sure, I'll take a look.

store_handler/eloq_data_store_service/eloq_store_data_store.cpp
```diff
@@ -164,7 +164,16 @@ void EloqStoreDataStore::Read(ReadRequest *read_req)
...
-    kv_read_req.SetArgs(eloq_store_table_id, key);
+    bool should_reopen = read_req->GetReopen();
+    if (should_reopen)
+    {
+        kv_read_req.SetArgs(eloq_store_table_id, std::string(key));
+    }
+    else
+    {
+        kv_read_req.SetArgs(eloq_store_table_id, key);
+    }
```
liangjchen: why does the reopen flag change how the kv request is set? the ``key`` is a string view comes from ``ReadRequest``. That does not change.

tx_service/include/cc/object_cc_map.h
liangjchen, line 1770: Could you please educate me on when is ``UploadTxCommandsCc `` used? It looks like it assumes the key's cc entry must exist and it knows the memory pointer.

tx_service/include/cc/object_cc_map.h
```diff
@@ -2065,6 +2081,23 @@ class ObjectCcMap : public TemplateCcMap<KeyT, ValueT, false, false>
                     // applied and there is no pending command.
                     cce->RecycleKeyLock(*shard_);
                 }
+                else
+                {
+#ifdef DATA_STORE_TYPE_ELOQDSS_ELOQSTORE
+                    const KeyT *key_ptr = ccp->KeyOfEntry(cce);
```
liangjchen: you don't need ``KeyOfEntry``. there is a local variable ``look_key``.

tx_service/include/cc/object_cc_map.h
```diff
@@ -2065,6 +2081,23 @@ class ObjectCcMap : public TemplateCcMap<KeyT, ValueT, false, false>
...
+                    int32_t part_id =
+                        Sharder::MapKeyHashToHashPartitionId(key_ptr->Hash());
+                    int64_t ng_term = Sharder::Instance().StandbyNodeTerm();
+                    shard_->FetchRecord(this->table_name_,
+                                        this->GetTableSchema(),
+                                        TxKey(key_ptr).Clone(),
```
liangjchen: you don't need to clone. when this key is passed to the data store client, the client will serialize to ``std::string`` before sending to the data store (even if it's local). -- Please verify.

tx_service/include/cc/object_cc_map.h
thweetkomputer: UploadTxCommandsCc is sent by a committing tx's coordinator during bucket migration: besides the normal write on the old owner, the tx's commands are also uploaded to the bucket's new owner (see tx_execution.cpp, "upload commands to the key bucket's new owner").  The cce pointer is safe: it comes from this tx's own forwarded AcquireWrite, which created the entry on the new owner and took the write lock. The lock is held until this request releases it at the end of Execute, and an entry with a held lock, buffered commands, or a pin can never be recycled (IsFree() requires the lock/extra-data structure to be empty).

tx_service/include/cc/object_cc_map.h
```diff
@@ -1821,9 +1821,25 @@ class ObjectCcMap : public TemplateCcMap<KeyT, ValueT, false, false>
...
+#ifdef DATA_STORE_TYPE_ELOQDSS_ELOQSTORE
+            if (!buffered_cmd_list.Empty())
+            {
+                const KeyT *key_ptr = ccp->KeyOfEntry(cce);
+                int32_t part_id =
+                    Sharder::MapKeyHashToHashPartitionId(key_ptr->Hash());
+                int64_t ng_term = Sharder::Instance().StandbyNodeTerm();
+                shard_->FetchRecord(this->table_name_,
+                                    this->GetTableSchema(),
+                                    TxKey(key_ptr).Clone(),
```
liangjchen: Remove ``Clone()``.

tx_service/include/cc/object_cc_map.h
```diff
@@ -2468,6 +2500,23 @@ class ObjectCcMap : public TemplateCcMap<KeyT, ValueT, false, false>
                 }
                 (void) lock_recycled;
             }
+#ifdef DATA_STORE_TYPE_ELOQDSS_ELOQSTORE
+            else if (buffered_cmd_list != nullptr)
+            {
+                const KeyT *key_ptr = ccp->KeyOfEntry(cce);
+                int32_t part_id =
+                    Sharder::MapKeyHashToHashPartitionId(key_ptr->Hash());
```
liangjchen: 1. there are local variables ``key`` and ``key_hash``. you don't need to re-compute the two. 
2. there is another piece of code in this function calling ``FetchRecord``. Consider using the lambda.

tx_service/include/cc/object_cc_map.h
```diff
@@ -2065,6 +2081,22 @@ class ObjectCcMap : public TemplateCcMap<KeyT, ValueT, false, false>
...
+                        Sharder::MapKeyHashToHashPartitionId(look_key->Hash());
+                    int64_t ng_term = Sharder::Instance().StandbyNodeTerm();
+                    shard_->FetchRecord(this->table_name_,
+                                        this->GetTableSchema(),
+                                        TxKey(look_key),
+                                        cce,
+                                        this->cc_ng_id_,
+                                        ng_term,
+                                        nullptr,  // requester
+                                        part_id);
```
liangjchen: 1. there is another call of ``FetchRecord`` in this function. Consider using a lambda. 
2. The prior call of ``FetchRecord`` uses ``req.StandbyNodeTerm()``. And you use ``Sharder::Instance().StandbyNodeTerm()``. Could you please verify if the two are equivalent? (I think so. Otherwise, the commands are forwarded to the wrong standby)

tx_service/src/cc/cc_req_misc.cpp
```diff
@@ -860,11 +862,15 @@ bool FetchRecordCc::Execute(CcShard &ccs)
...
+#ifdef DATA_STORE_TYPE_ELOQDSS_ELOQSTORE
+    bool should_reopen = false;
+    TxKey reopen_key;
+#endif
+
     if (lock_->GetCcEntry() != nullptr)
     {
         assert(lock_->GetCcMap() != nullptr);
         assert(lock_->GetCcEntry() == cce_);
-        // if the referenced cce is already invalid, we do not need to care
```
liangjchen, line 867: Is this line deleted by accident?

tx_service/src/cc/cc_req_misc.cpp
```diff
@@ -918,9 +924,45 @@ bool FetchRecordCc::Execute(CcShard &ccs)
                 }
             }
         }
+
+#ifdef DATA_STORE_TYPE_ELOQDSS_ELOQSTORE
+        should_reopen = error_code_ == 0 && cce_->HasBufferedCommandList();
```
liangjchen: I think something important is missing. 

``HasBufferedCommandList()`` does not imply reopen. Reopen is needed after the fetched record is back filled, all pending commands are evaluated, and there are still pending commands after evaluation. What's missing here is that backfilling, which is done at ``lock_->GetCcMap()->BackFill()``, does not trigger evaluating pending commands. So, we need to first override ``BackFill`` for ``ObjectCcMap`` to process pending commands.

liangjchen: A more evolved thinking needs to consider the possibility that evaluating pending commands may monopolize  CPU and thus may need to yield. A stackful (boost::context) or stackless (std::generator) coroutine implementation is much easier to achieve this than the current ``FetchRecordCc::Execute``, which is essentially a handmade state machine. Let's leave this for future work.

tx_service/src/cc/cc_req_misc.cpp
```diff
@@ -918,9 +924,45 @@ bool FetchRecordCc::Execute(CcShard &ccs)
...
+        if (should_reopen)
+        {
+            reopen_key = tx_key_.Clone();
```
liangjchen: For ``FetchRecord``, input keys do not need ``Clone()``.

thweetkomputer: I removed `Clone()` at the other two `FetchRecord` call sites (`object_cc_map.h` in `Execute(UploadTxCommandsCc&)` and `Execute(ReplayLogCc&)`) — you're right there, and `DataStoreServiceClient` promotes a non-owning key to owning anyway (`data_store_service_client.cpp:4536` / `:4650`), so the caller-side clone is redundant.

But I'd like to keep it at this one, because here the key is not passed straight into `FetchRecord` — it is stashed across the destruction of `this`:

```cpp
reopen_key = tx_key_.Clone();      // 933
...
ccs.RemoveFetchRecordRequest(cce_); // 947  -> fetch_record_reqs_.erase(cce) -> destroys *this, incl. tx_key_
...
ccs.FetchRecord(..., std::move(reopen_key), ..., /*reopen*/ true); // 952
```

`fetch_record_reqs_` is `std::unordered_map<LruEntry *, FetchRecordCc>` (`cc_shard.h:1283`), so the `erase` runs `~FetchRecordCc()`. That is also why lines 939-944 copy `table_name_` / `table_schema_` / `cce_` / `cc_ng_id_` / `cc_ng_term_` / `partition_id_` into locals first.

And the reopen fetch has to come *after* the erase: `CcShard::FetchRecord` coalesces by `LruEntry *` via `try_emplace`, so if the current (dying) request were still in the map, `try_emplace` would hand back the existing entry, `RequesterCount()` would not be 1, and the reopen fetch would never be dispatched — that was the ordering fix for the earlier review comment on this function.

So `tx_key_` is already gone by the time the reopen fetch runs; dropping the `Clone()` here would be a use-after-free.

Two follow-ups, happy to take either:

1. `reopen_key = std::move(tx_key_)` would save the copy. It is safe as long as `tx_key_` is guaranteed to be owning at this point — it is on the DSS path because of the promote above, but I have not convinced myself about the `fetch_from_primary_` path. Do you know offhand?
2. Cleaner still: give `FetchRecordCc` a `Reset()` (like `FetchSnapshotCc::Reset`) and re-arm it in place with `reopen = true` instead of erase + re-emplace. That removes both the local-copy dance and the clone. Since it touches the same code as the `ObjectCcMap::BackFill` override you asked for, I'd rather do the two together in a follow-up — let me know if you'd prefer it in this PR.

tx_service/src/cc/cc_req_misc.cpp
thweetkomputer: object_cc_map has overrided BackFill to consume buffered commands.

tx_service/src/cc/cc_req_misc.cpp
thweetkomputer: yes, this line is added back.

liangjchen (approved):

PR #491, opened 2026-06-12
fix: avoid bthread mutex deadlock between tx processors and bthread waiters

liunyl:
## Problem

Same bug class as the CkptTsCc hang fixed in #483, found in several more CC requests after the issue reappeared in the field.

A `bthread::Mutex`/`ConditionVariable` shared between `CcRequest::Execute()` and a bthread owner can deadlock under our brpc fork:

1. `Execute()` runs on the brpc worker **main stack** (via `ProcessModulesTask()` → `TxProcessor::RunOneRound()`, with a fixed 1:1 worker↔shard binding), so a contended `lock()` there parks the worker as a **pthread waiter** in the mutex's butex queue.
2. The owner bthread waits on the same mutex/cv. `butex_wake` prefers bound-bthread waiters, and bound bthreads can only run on their home worker (`_bound_rq` is never stolen).
3. When an unlock routes the wake to a bthread bound to the blocked worker, the bthread sits in that worker's bound run queue forever while the worker sleeps in `butex_wait`. The mutex may even end up free — nobody can take it. `wait_for` timeouts don't help: in the end state no timer is pending.

The dangerous shape is **concurrent multi-acquisition** (a request broadcast to multiple cores, or one mutex shared by several in-flight requests) combined with a bthread waiter. Single-flight/single-waiter requests are structurally safe.

## Fix

Convert the affected requests to the pattern used by the CkptTsCc fix (#483): shard-side completion is tracked with atomics, the waiter polls with `bthread_usleep` exponential backoff (100us → 100ms), and no bthread::Mutex is shared between tx processors and bthreads.

| Request | Change |
|---|---|
| `ActiveTxMaxTsCc` (used by `ResetStandbySequenceId` RPC) | atomic countdown + polling `Wait()` |
| `WaitableCc` (covers `OnLeaderStart` / `SubscribePrimaryNode` / `UpdateInMemoryClusterConfig` / standby RPC handlers, etc.) | blocking `Wait()` polls; `IsFinished/IsError/ErrorCode` read atomics; `mux_` retained **only** for the coroutine yield/resume handshake whose waiter is a flush-worker `std::thread` |
| `ClearCcNodeGroup` | atomic counter + `done_` flag set after the last core's catalog/range/bucket cleanup |
| `EscalateStandbyCcmCc` | atomic countdown + polling `Wait()` |
| `ClearTxCc` | atomic countdown + polling `Wait()` (was `std::mutex`; safe today since the waiter is the recovery `std::thread`, but a latent landmine for any future bthread caller) |
| `DbSizeCc` / `RemoteDbSizeCc` | atomic ref counts; `Wait()` keeps the give-up-on-remote-after-2s semantics; `remote_ref_cnt_` decremented before `total_ref_cnt_` so the give-up condition never fires while a local core is pending |
| `UploadBatchCc` + `CcNodeService::UploadBatch` + sk_generator upload path | external coordination state (`req_mux/req_cv/finished_req_cnt/req_result/ng_term`) replaced with `std::atomic` references; `ValidTermCheck()` uses CAS; the `UploadBatchClosure` callback (runs on a bthread) updates atomics lock-free |

Left intentionally unchanged (structurally safe): `CheckTxStatusCc`, `WaitNoNakedBucketRefCc`, `UploadRangeSlicesCc`, `UploadBatchSlicesCc` (single-flight + single waiter), `UpdateCceCkptTsCc` and the flush-data coroutine path (waiter is a dedicated `std::thread`).

## Verification

- Full Debug build of eloqkv (with this tx_service) passes: `install/build.sh` exit 0.
- Recommended runtime coverage: standby subscribe/resubscribe, leader failover/step-down, `UploadBatch` during bucket migration, and the Redis `dbsize` command — these are the paths that previously hung.

## Summary by CodeRabbit

* **Refactor**
  * Reworked internal completion/error coordination across transaction-control flows to use atomic, lock-free tracking instead of lock/condition-variable synchronization.
  * Updated wait behavior to use bounded exponential-backoff polling, improving stability and avoiding cross-context wait issues.
  * Added explicit `Wait()` support for clearer completion handling in CC request paths.
* **Tests**
  * Added coverage for wait completion behavior, error latching, and repeated concurrent completion rounds.

### DIFF
--- a/tx_service/include/cc/cc_req_misc.h
+++ b/tx_service/include/cc/cc_req_misc.h
@@ -339,23 +339,19 @@ public:
 
     bool Execute(CcShard &ccs) override;
 
-    void Wait()
-    {
-        std::unique_lock lk(mux_);
-        while (finish_cnt_ != core_cnt_)
-        {
-            wait_cv_.wait(lk);
-        }
-    }
+    // ClearCcNodeGroup is issued and Waited on RPC handler bthread, while
+    // Execute() runs on tx processors (the brpc worker main stack), so the
+    // two sides must not share a bthread::Mutex. Wait() polls an atomic
+    // flag instead.
+    void Wait();
 
 private:
     const uint32_t cc_ng_id_;
     const uint16_t core_cnt_;
-    uint16_t finish_cnt_{0};
-    // ClearCcNodeGroup is issued and Waited on RPC handler bthread, use bthread
-    // mutex and condition variable.
-    bthread::Mutex mux_;
-    bthread::ConditionVariable wait_cv_;
+    std::atomic<uint16_t> finish_cnt_{0};
+    // Set by the last core after it has dropped the node group's table
+    // statistics, catalogs, ranges and bucket info.
+    std::atomic<bool> done_{false};
 };
 
 struct FillStoreSliceCc;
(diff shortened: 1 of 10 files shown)


liunyl:
@coderabbit review

thweetkomputer (approved):

PR #492, opened 2026-06-12
fix: avoid bthread mutex waits in shard completion paths

starrysky9959:
## Summary
- cherry-pick release branch fix  onto 
- avoid waiting on bthread mutexes in shard completion paths
- keep  aligned with the fix already merged into 

## Testing
- not run

## Summary by CodeRabbit

* **Performance Improvements**
  * Replaced heavy lock-based coordination with atomic counters and backoff polling, improving responsiveness and reducing lock contention in transaction processing.
* **Bug Fixes / Reliability**
  * More robust completion and abort handling with atomic state updates and timeout-aware wait logic to reduce stalls and spurious waits.

### DIFF
--- a/tx_service/include/cc/cc_req_misc.h
+++ b/tx_service/include/cc/cc_req_misc.h
@@ -871,11 +871,11 @@ public:
     void Reset(std::function<bool(CcShard &ccs)> task = {},
                uint16_t core_cnt = 1)
     {
-        std::lock_guard<bthread::Mutex> lk(mux_);
         RunOnTxProcessorCc::Reset(std::move(task));
 
-        unfinished_cnt_ = core_cnt;
-        error_code_ = CcErrorCode::NO_ERROR;
+        unfinished_cnt_.store(core_cnt, std::memory_order_relaxed);
+        error_code_.store(CcErrorCode::NO_ERROR, std::memory_order_relaxed);
+        waiting_.store(false, std::memory_order_relaxed);
     }
 
     void SetCoroCallbacks(const std::function<void()> *yield_fn,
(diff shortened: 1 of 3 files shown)


tx_service/include/cc/cc_req_misc.h
```diff
@@ -902,53 +905,47 @@ struct WaitableCc : public RunOnTxProcessorCc
...
         }
-        std::unique_lock<bthread::Mutex> lk(mux_);
-        while (unfinished_cnt_)
+        while (unfinished_cnt_.load(std::memory_order_acquire) > 0)
         {
             waiting_.store(true, std::memory_order_release);
-            lk.unlock();
+            if (unfinished_cnt_.load(std::memory_order_acquire) == 0)
+            {
+                waiting_.store(false, std::memory_order_release);
```
thweetkomputer, line 913: The waiter enters the loop after observing unfinished_cnt_ == 1 (W1) and sets waiting_ = true (W2). At this point the completer on the TxProcessor thread decrements the counter to zero (C1) and then exchanges waiting_ from true to false (C2) — since it observed true, it commits to calling resume_fn. Meanwhile, the waiter's re-check (W3) observes unfinished_cnt_ == 0 and takes the early-exit branch: it stores waiting_ = false (W4, which is now a no-op because C2 already cleared the flag) and breaks out of the loop without ever calling yield_fn (W5). Finally the completer invokes resume_fn() (C3), which unconditionally enqueues the coroutine's CoroCtx into resume_queue_.
▎
▎ The result is an orphan resume: a resume has been published with no matching yield to consume it. resume_fn is not a level-triggered notification like a condition variable — every enqueued entry will eventually be popped by the flush worker, which calls ctx->coro_.resume(). Since the coroutine never suspended (and may have already run to completion by then), the worker ends up resuming a running or finished boost::context continuation, which is undefined behavior.
▎
▎ Note the bug is timing-dependent: if W4 wins the race against C2 instead, the completer's exchange reads false, no resume is published, and everything works — the failure window is only the gap between W2 and W4.

tx_service/include/cc/cc_request.h
```diff
@@ -8433,13 +8427,7 @@ struct DbSizeCc : public CcRequestBase
...
         }
 
-        std::unique_lock lk(mux_);
-        if (--total_ref_cnt_ == 0)
-        {
-            cv_.notify_one();
-        }
-
-        return false;
+        return OnLocalRefFinished();
```
thweetkomputer, line 8430: this should always return false? or DbSizeCc is recycled and reused by another caller.

OnLocalRefFinished();
return false;

Issue #476, opened 2026-04-21
[Feature]: Standby nodes periodically trigger refresh and reopen with newest manifset.


thweetkomputer:
- **Summary (Provide a general summary of the issue in the Title above):**(required)

- [ ] Standby nodes should periodically trigger refresh in case that master node disconnect to it.

- [ ] Master nodes do not snapshot to let standby nodes reopen with snapshots.

- [ ] ccmap should try to consume buffered commands periodically to avoid memory occupation.

- **Detailed Description (Provide a detailed description of the change or addition you are proposing):**(required)

- **Software version (the version of the software you're running):**(required)
e.g. 0.0.1
- **Possible Implementation(suggest an idea for implementing addition or change)**(optional)

PR #521, opened 2026-07-11
Fix DSS drain gates

thweetkomputer:
## Summary

This PR closes two DataStoreService shutdown/drain races:

- Prevents a write request from being accepted concurrently with `ReadWrite -> ReadOnly` while the drainer observes zero in-flight writes.
- Prevents new read/scan submissions from entering the backend after `ReadOnly -> Closed` begins draining, which is required for EloqStore because `Stop()` must not race with a later `ExecAsyn()` submission.

It also advances the nested `eloqstore` submodule to the latest `main` commit:

- `acd3cb3 fix: complete requests racing Stop() instead of crashing or hanging (#483)`
- includes the recent fixes from #480, #476, and #469.

### Write drain gate

The write path already increments `ongoing_write_requests_` before checking shard status. However, the previous ordering used release/acquire operations, which does not impose one global order across the writer's count increment/status load and the drainer's status CAS/count load.

This PR changes the write gate operations to `seq_cst`:

- write request count increment/decrement
- post-increment shard status loads on write paths
- `ReadWrite -> ReadOnly` CAS
- write-drain count load

With one sequentially consistent order, the unsafe outcome where a writer still observes `ReadWrite` while the drainer already observes count `0` is prevented.

### Read/scan submit drain gate

Read and scan paths now enter a per-thread submit window before loading shard status. `ReadOnly -> Closed` first switches the shard status to `Starting`, then waits for all submit windows to drain before calling backend `Shutdown()`.

This blocks the problematic interleaving:

1. a read/scan request is in DataStoreService and about to submit to the backend;
2. close switches the shard to closing and drains;
3. backend shutdown/stop runs;
4. the read/scan request submits after shutdown has begun.

The submit window is intentionally scoped only around the status check and backend submission. It is not a full async request lifetime counter.

That is deliberate for EloqStore: the bug is submission-after-stop, not waiting for every accepted read callback before close. Extending the counter to `SetFinish()` would add a cross-thread atomic decrement to the hot read path.

For RocksDB/RocksDBCloud, accepted reads/scans are already safe because their async behavior is implemented by DataStoreService's `query_worker_pool_`. `Shutdown()` joins the worker pool, and workers drain queued work before exiting, so the DB is closed/deleted only after accepted worker tasks complete.

## Validation

- `cmake --build bld-eloqstore-local --target eloqkv -j2`
- `git diff --check -- store_handler/eloq_data_store_service/data_store_service.cpp store_handler/eloq_data_store_service/data_store_service.h`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of read and scan operations during datastore shutdown by tracking and draining in-flight read activity before shutdown completes.
  * Prevented shutdown from interrupting in-progress read requests, scans, and scan-close flows.
  * Strengthened synchronization around shard status handling to reduce race conditions under concurrent load.
* **Chores**
  * Updated the checked-in datastore service submodule reference.

### DIFF
--- a/store_handler/eloq_data_store_service/data_store_service.cpp
+++ b/store_handler/eloq_data_store_service/data_store_service.cpp
@@ -85,6 +85,11 @@ thread_local ObjectPool<CreateSnapshotForBackupLocalRequest>
 thread_local ObjectPool<SyncFileCacheLocalRequest>
     local_sync_file_cache_req_pool_;
 
+thread_local const DataStoreService *tls_read_submit_slot_owner = nullptr;
+thread_local uint32_t tls_read_submit_slot_idx = UINT32_MAX;
+thread_local std::array<uint32_t, DataStoreService::kMaxShardCount>
+    tls_read_submit_depths{};
+
 TTLWrapperCache::TTLWrapperCache()
 {
     ttl_check_running_ = true;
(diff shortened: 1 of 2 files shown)


thweetkomputer:
I have performed a benchmark on an 8-core ARM VM. No degradation is found.

liunyl (approved):

PR #522, opened 2026-07-11
Fix schema-op fallthrough and command ID truncation

### LINKED ISSUE #500: UpsertTableOp::Forward: missing return after ForceToFinish → use-after-free of pooled schema_op_ and finished txm
In the schema-change state machine, the `unlock_cluster_config_op_` error branch calls `ForceToFinish(txm)` (which finishes the txm and returns `schema_op_` to the shared pool) but then **falls through** and keeps using `schema_op_` / the txm — pushing and processing `upsert_kv_table_op_` on an object that may already have been recycled and handed to another transaction.

## Evidence

`tx_service/src/tx_operation.cpp:2729-2756` (`UpsertTableOp::Forward`):
```cpp
else if (op_ == &unlock_cluster_config_op_) {
    ...
    if (unlock_cluster_config_op_.hd_result_.IsError()) {
        if (txm->CheckLeaderTerm()) { assert(false); }
        else { ForceToFinish(txm); }     // finishes txm, recycles schema_op_  -- no return!
    }
    op_ = &upsert_kv_table_op_;           // continues to use the (possibly recycled) op
    upsert_kv_table_op_.table_schema_old_ = catalog_rec_.Schema();
    ...
    txm->PushOperation(&upsert_kv_table_op_);
    txm->Process(upsert_kv_table_op_);    // operate on a finished txm
}
```
Every other `ForceToFinish` site in this function returns immediately; this branch is missing the `return`. A sibling release-build fallthrough exists in `UpsertTableIndexOp` (`tx_index_operation.cpp:459-503`).

## Scenario
A `CreateTable`/`Update`/`TruncateTable` whose `unlock_cluster_config_op_` (a local read-lock release) reports an error precisely when this node is no longer leader (the only branch that hits `ForceToFinish`): the txm is finished and `schema_op_` returned to the pool, then the code pushes `upsert_kv_table_op_` and processes it → operating on a recycled op / finished txm. If the pool slot was reassigned, this corrupts another in-flight schema transaction.

**Fix:** `return` after `ForceToFinish(txm)` in this branch (and audit the `UpsertTableIndexOp` sibling).

---
Found during a code audit (PR #493). Verified against source at the cited lines.

🤖 Found with [Claude Code](https://claude.com/claude-code)

### LINKED ISSUE #502: command_id_ is uint16 behind a uint32 wire field → wraparound can accept a stale remote response in long transactions
A transaction's `command_id_` — the per-tx counter that, together with the constant `tx_number`, is the sole guard against accepting a stale/duplicate remote cc response — is stored as 16 bits while the wire field is 32 bits. A transaction issuing more than 65536 operations/remote sends wraps the counter, after which a sufficiently delayed duplicate response can alias the current op and be accepted, writing a stale `CcEntryAddr`/`commit_ts` into the handler result.

## Evidence

`tx_service/include/tx_execution.h:595`:
```cpp
std::atomic<uint16_t> command_id_;   // reset to 0 per tx; ++ on each request / remote send
```
Wire field is 32-bit: `tx_service/include/proto/cc_request.proto:1127, 1462` (`uint32 command_id`).
Response matching uses it as the staleness guard at e.g. `tx_service/src/remote/cc_stream_receiver.cpp:352, 477, 588, 732, 789, 865`:
```cpp
txm->CommandId() != msg->command_id() || hd_res->IsFinished()
```
`command_id_` is reset to 0 only per-tx (`tx_execution.cpp:140`) and incremented at many sites (`:716, 1891, 2317, 2438, 2893, 3929, ...`). `tx_number` is constant for the life of a (possibly long) EloqKV session transaction.

## Scenario
A long-lived interactive `BEGIN … COMMIT` session (or a very large MULTI/Lua) performs > 65536 operations; `command_id_` wraps. A remote cc response from ~65536 commands earlier that is still in flight has an id that now equals the current `command_id_`; with the same `tx_number` it passes the guard and is applied to the wrong operation → wrong lock release / stale read result.

Low real-world probability (needs a very long tx plus a very stale in-flight response), but a genuine latent correctness bug. **Fix is free:** widen `command_id_` to `uint32_t` to match the wire field.

---
Found during a code audit (PR #493). Verified against source at the cited lines.

🤖 Found with [Claude Code](https://claude.com/claude-code)

liunyl:
## Summary

- return immediately after `ForceToFinish` in `UpsertTableOp::Forward`
- widen transaction command IDs and CC handler boundaries to `uint32_t`, matching the protobuf wire field
- add a compile-time regression assertion for the public command ID type

## Root cause

The schema operation continued after its transaction and pooled operation state were finished. Separately, the per-transaction command counter and CC handler parameters were narrower than the 32-bit wire field, allowing wraparound/truncation to alias stale remote responses.

Fixes #500
Fixes #502

## Validation

- `clang-format` 18.1.8
- `git diff --check`
- full local configure blocked by missing prebuilt RocksDB headers; CI will run the repository build and tests


tx_service/include/tx_execution.h
```diff
@@ -212,7 +212,7 @@ class TransactionExecution : public LinkedTransaction
 
     int64_t TxTerm() const;
 
-    uint16_t CommandId() const;
+    uint32_t CommandId() const;
```
liunyl, line 215: [P1] Widen every local `CommandId()` consumer too. `TransactionExecution::Forward()` still stores this value in `uint16_t cmd_id` (`tx_execution.cpp:471`). At command 65,536 that local becomes 0 while `CommandId()` is 65,536, so the same operation can no longer satisfy the loop's `cmd_id == CommandId()` break condition and is forwarded repeatedly. The EXT_TX_PROC progress paths also narrow at `tx_service.h:738`, `:762`, `:818`, `:1024`, and `:1029`. Change these tracking values to `uint32_t` and add a runtime boundary check; the new `static_assert` only verifies this return type and misses the truncating callers.

liunyl (commented):
Found one blocking correctness issue: the command-ID widening stops before all local consumers, creating a new failure once the ID exceeds 16 bits.

PR #523, opened 2026-07-11
fix: remote-owned keys forwarded/logged pre-ExecuteOn command images (eloqkv#509)

liunyl:
## Summary

Fixes the engine side of eloqdata/eloqkv#509: for a key owned by a **remote** node group, both the standby forward and the coordinator's WAL captured the **pre-`ExecuteOn`** command image, while standby apply and log replay are commit-only (never re-run `ExecuteOn`) — so commands whose `ExecuteOn` mutates the command in place (EloqKV ZADD NX/XX/GT/LT filtering, LTRIM index normalization, SPOP member selection, EXPIRE-family TTL snapshots) diverged replicas and recovery.

### Fix 1 — standby forward serializes the executed command (`83748de`)

`StandbyForwardEntry::AddTxCommand` now serializes `cc_req.GetCommand()` (the owner-side executed command) for **both** local and remote `ApplyCc`, instead of copying the coordinator's original bytes in the remote branch. Audit confirmed no other consumer forwards/persists the pre-`ExecuteOn` image for replay.

### Fix 2 — owner TTL facts round-trip in `ApplyResponse`; expired-recreation ttl rule fix (`3c197b5`)

`ApplyResponse` gains four fields (proto3, rolling-upgrade safe by default values):

| field | purpose |
|---|---|
| `ttl_reset = 10` | owner reset a live TTL → coordinator must WAL-log the full-object snapshot |
| `recover_cmd_image = 11` | the owner-side `RecoverObjectCommand` snapshot image (set only with `ttl_reset`) |
| `ttl = 12` | owner-computed **post-command** object ttl → the WAL record's validity horizon (was: remote always logged `UINT64_MAX`, letting recovery resurrect expired objects) |
| `ttl_expired = 13` | command consumed an expired key → coordinator prepends the retire `DelCommand` overwrite record |

Plumbing: `FillApplyResponse` (owner, extracted from `RemoteApplyCc::post_lambda_`) / `BackfillObjectCommandResult` (coordinator, extracted from `cc_stream_receiver`) — behavior-preserving extractions that also make the boundary unit-testable. The snapshot image is logged through a new `CmdSetEntry::AddOverwriteCommandImage`, which preserves the overwrite invariants (`cmd_str_list_` cleared, `ignore_previous_version_`/`has_overwrite` set, ttl updated). An old owner omitting the fields degrades to today's behavior (`ttl == 0 → UINT64_MAX` fallback).

**Pre-existing bug fixed on the way (bites purely LOCAL keys too):** the owner reported the **pre-command** (already expired) ttl for a write that consumed an expired key and recreated it; the WAL record then carried an expired validity horizon and **recovery discarded the acknowledged recreation** (`ReplayLogCc` treats an expired `valid_scope` as "object deleted"). The extracted rule `ComputeReportedTtl` now reports the post-command ttl (`UINT64_MAX` for a fresh recreation).

### Tests (red-first — each failure reproduced on the pre-fix code and recorded)

- `StandbyForward-Test`: remote case forwarded the pre-image (`"PRE" == "POST"` assertion failure pre-fix); local green guard.
- `ApplyResponseBoundary-Test`: fill → real protobuf wire round-trip → resend equivalence → backfill → `CmdSetEntry` WAL-input legs. Red pre-fix: snapshot image dropped; finite ttl dropped (resurrection horizon); `ComputeReportedTtl` returned the expired value. Green guards: overwrite invariants with a seeded entry; old-owner `ttl` fallback.
- Full `ctest`: 51 tests green (was 46 before the branch).

### Rolling upgrade / mixed versions

No wire-format break (proto3 additive fields; forwarded command images unchanged in format). During a mixed-version rollout no direction introduces new breakage; un-upgraded owners simply keep today's (buggy) behavior until upgraded — the fix takes full effect once coordinator and owner are both upgraded. If standby divergence from the pre-fix period is suspected, a standby re-sync after upgrade is advisable.

### Known follow-ups (out of scope, same bug class)

- The **multi-object** post-process path has the `ttl_expired` retire block but no `ttl_reset` snapshot logging (no current multi-object command both resets a TTL and is non-overwrite, so theoretical today).
- No e2e harness exists for remote-owner + standby + WAL-replay scenarios (TestCluster has a record-only workload, no standbys, skip_wal recovery); unit-level boundary tests above are the current guard.

Companion EloqKV PR: commit-side replay determinism for ZADD/LTRIM (+ docs); the eloqkv submodule pointer bump follows this PR's merge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction durability and recovery for remote commands that change or reset TTL values.
  * Ensured standby replication forwards the command state after execution, improving consistency.
  * Preserved correct TTL behavior for live, expired, recreated, and legacy objects.

* **Tests**
  * Added coverage for remote apply responses, TTL recovery, overwrite records, and standby forwarding behavior.

* **Documentation**
  * Clarified WAL recovery, TTL handling, and standby replication behavior.


tx_service/include/command_set.h
```diff
@@ -123,6 +123,72 @@ class CommandSet
         }
     }
 
+    // Sibling of AddObjectCommand that logs a pre-serialized full-object
+    // snapshot image (from a remote owner) as an overwrite record instead of
+    // serializing a coordinator-held command. Entry lookup and bookkeeping are
+    // identical; only the command payload differs.
+    void AddObjectCommandImage(const TableName &table_name,
```
liunyl, line 98: [P3] Share the existing command-entry setup. This method duplicates `AddObjectCommand`'s table lookup, entry creation, version/lock bookkeeping, write-lock count, and forwarding setup; the only new behavior is `entry.AddOverwriteCommandImage(...)`. Put the shared get/update logic in one small path and let both entry points select only how the command is appended. That removes roughly 40 duplicated lines and prevents the two bookkeeping paths from drifting.

liunyl (commented):
No blocking correctness findings in the standby-forward / ApplyResponse / WAL paths reviewed. One non-blocking simplification would keep the new path leaner.

tx_service/include/command_set.h
liunyl: Done in the latest push: extracted `UpsertObjectCmdEntry` (lookup/create + version/lock/vali refresh) and `SetupForwardEntry` as private helpers; `AddObjectCommand` and `AddObjectCommandImage` now differ only in how the payload is appended (`AddCommand` vs `AddOverwriteCommandImage`). ctest 51/51 green after the refactor.

liunyl:
Review feedback addressed in the latest push (`dde2e3b`):

- **[P3] command_set.h duplication** — extracted `UpsertObjectCmdEntry` + `SetupForwardEntry`; the two entry points now share one bookkeeping path and differ only in the append call. (Replied inline.)
- **cpplint** — added the missing `<utility>`/`<memory>` (command_set.h) and `<string>` (remote_cc_request.cpp) includes; reordered C++ system headers before catch2 in the two new test files.
- **CLA / commit author** — the branch commits were authored with an unlinked local identity (`Ubuntu <ubuntu@...internal>`); rewritten to `liunyl <lukeliu970702@gmail.com>` and force-pushed, so the CLA check can associate them once the CLA is signed.

Engine `ctest` 51/51 green after the refactor.

thweetkomputer (approved):

MrGuin (approved):

liunyl:
@coderabbitai review

PR #526, opened 2026-07-14
fix: retain scan reads until transaction finalization

### LINKED ISSUE #508: Use-after-free: PostReadCc dereferences a recycled NonBlockingLock when a read-intent is released before its post-read (under CleanDataForTest kickout)
## Summary

A pre-existing **use-after-free** in the cc-entry lock lifetime. Under a multi-node cluster with the data store + aggressive eviction (`kickout_data_for_test=true`, `checkpointer_interval=1`) plus concurrent scans/writes, a `PostReadCc` can dereference a `NonBlockingLock` that has already been recycled, aborting the node.

## Crash signature

```
eloqkv: .../abseil-cpp/absl/container/internal/raw_hash_set.h:325:
probe_seq<16>::probe_seq(size_t, size_t): Assertion `((mask + 1) & mask) == 0 && "not a mask"' failed.   (SIGABRT)
```

Crashing thread (core + gdb):

```
#3  absl::container_internal::probe_seq<16>::probe_seq        (corrupt capacity)
#6  raw_hash_set<FlatHashMapPolicy<uint64,uint32>>::find_non_soo
#8  txservice::NonBlockingLock::ReleaseReadIntent             non_blocking_lock.cpp:503
#9  txservice::NonBlockingLock::ClearTx                       non_blocking_lock.cpp:562
#10 txservice::CcMap::ReleaseCceLock                          cc_map.cpp:412  (lk_type=NoLock, recycle_lock=true)
#11 TemplateCcMap<EloqKey,RedisEloqObject>::Execute           template_cc_map.h:1314
#12 PostReadCc::Execute                                       cc_request.h:262
#13 CcShard::ProcessRequests -> TxProcessor::RunOneRound      (is_ext_proc=true)
```

In the core, the `NonBlockingLock`'s `read_intentions_` (`absl::flat_hash_map<TxNumber,uint32_t>`) has garbage `capacity_` / `size_` / slot pointer — the lock object was recycled/clobbered while still referenced.

## Root cause

`PostReadCc::Execute` (`template_cc_map.h:1243`) obtains the entry from the cached `cce_addr` (`cce_addr.ExtractCce()`, a `KeyGapLockAndExtraData*` captured at read time), then calls `cc_entry.GetKeyLock()->ReleaseReadIntent()`. If the cce's read-intent is released **before** that `PostReadCc` runs, the entry becomes `IsFree()` (`cc_lock_and_extra_ == nullptr && IsPersistent()`), the `CleanDataForTest` kickout (`LocalCcShards::KickoutDataForTest` → `KickoutCcEntryCc` with `CleanType::CleanDataForTest`) recycles it, and the later `PostReadCc` dereferences the recycled lock → UAF.

## What is already correct (so the gap is the early intent release)

- Read-set insertion already guards `lock_type != LockType::NoLock` at every site (point read `tx_execution.cpp:2206`; scan sites `tx_execution.cpp:3007 / 3148 / 3369 / 5886`). `ReadCommitted` reads map to `NoLock` (`LockTypeUtil::DeduceLockType`) and are not tracked; scans `AcquireReadIntent` per tuple.
- The kickout gate already skips locked entries: `CleanDataForTest` → `CcPageCleanGuardWithoutKickoutCc::CanBeCleaned` = `IsFree() && !GetBeingCkpt()`.

So the entry was `IsFree()` at kickout time, i.e. **its read-intent had been released before the post-read**. Suspected early-release sites (a read-intent refcount interaction): scan `DrainScanner` (`PostReadType::Release`, `tx_execution.cpp ~3651`), the transient last-cce pin `DecrReadIntent` (`template_cc_map.h:3884` / `~3891`), or a `DedupRead` edge (`tx_execution.cpp:6370`).

## Dead code noticed nearby

The "record-and-don't-free a still-locked cce" defense is not wired up: `CcPageCleanGuard::MarkClean(idx, delay_free)` ignores its `delay_free` argument, and `LruEntry::ClearLocks(ccs, ng, invalidate_owner_term)` never references `invalidate_owner_term` (no caller passes `true`) — vestigial since #168 ("Restructure for eloqconverge"). For the genuine stale-ex-leader post-read, the only active guard is the term check (`PostReadCc::CheckLeaderTerm` + `ReadWriteSet::ClearDataReadSet` read-lock-term validation), which fires only on a term change (failover) and does **not** cover the same-term kickout case above.

## Relation to PR #491 (bthread-mutex CC-request fix)

**Not a regression from #491.** Verified by A/B (built fix `@1448c93` vs pre-fix `@f62f0fa`): both pass `scan.tcl` alone 15× and both crash identically under concurrent-writer stress with the same `NonBlockingLock` assertion. The CI failure on the eloqkv PR (`eloqkv-pr-ent/compile-test` build #285, config 7 = 3-node cluster + eloqstore + `kickout_data_for_test`, `scan.tcl` first `flushdb` → `I/O error reading reply`) is an intermittent manifestation of this pre-existing race, not a new bug introduced by #491.

## Reproduction (flaky / timing-sensitive)

- Debug build, `WITH_DATA_STORE=ELOQDSS_ELOQSTORE`, `EXT_TX_PROC_ENABLED=ON`, `WITH_LOG_SERVICE=ON`, `OPEN_LOG_SERVICE=OFF`, `FORK_HM_PROCESS=ON`.
- 3-node local cluster, shared eloqstore via minio, `--core_number=2 --checkpointer_interval=1 --kickout_data_for_test=true --max_processing_time_microseconds=1000`.
- Concurrent writers (`SET` churn) + `scan.tcl` loop → a data node SIGABRTs intermittently (usually within 1–2 scan iterations under light load; heavier load tends to hang instead).
- A **deterministic** repro / regression test is best done with `CODE_FAULT_INJECTOR` (force a kickout/recycle of a read-set cce right before its `PostReadCc`).

## Proposed fix direction

Ensure a cce referenced by an in-flight `PostReadCc` (i.e. in a tx's data/scan read-set) retains its read-intent until the post-read releases it, so the kickout's `IsFree()` gate skips it. Audit the read-intent refcount accounting across the scan drain / transient last-cce pin / dedup paths. Optionally remove (or properly wire up) the dead `delay_free` / `invalidate_owner_term` machinery.

liunyl:
## Context

Fixes eloqdata/tx_service#508.

This PR fixes two scan-ownership defects:

1. Scan close/drain paths could release semantic reads and scanner-only pins before transaction commit or abort.
2. A later hash-scan batch ignored a completed CC-map memory source and scanned it again while store buckets were still unfinished. That rescan could materialize the same CCE again and add another `ReadIntent` that was not part of the intended batch ownership.

The second defect is not the previously proposed KV-callback/continuation race. A live in-batch continuation exists only while the memory pass is unfinished; `memory_scan_is_finished_` becomes true only after the pass reaches the CC-map end.

## Behavior before and after

Before:

- scan close/drain could remove read ownership while the transaction was still active;
- after a core completed its memory pass, a later client batch could enter that CC map again whenever at least one KV bucket was still unfinished;
- the redundant memory pass could add another `ReadIntent` and duplicate memory-source work.

After:

- semantic scan reads and scanner-only pins remain accounted for until transaction finalization;
- a core whose memory source is still marked finished starts any required KV fetches, skips the CC map, and adds no new memory-source `ReadIntent`;
- existing transaction-owned intents remain until commit or abort.

## Implementation

- Remove the scan-specific early-release operation and keep returned locking reads in the transaction read set.
- Consolidate remaining, trailing, range-resume, and error-path scanner ownership in `DrainScanner`, using release-only entries for implementation pins that must not participate in OCC version validation.
- Preserve transaction-lifetime backing for scan table-name views instead of duplicating table-name state.
- In both local and remote hash-scan `Execute` paths:
  - keep term validation and the original terminal checks at the top;
  - launch fetches for unfinished KV buckets as before;
  - on a first-entry path with a finished memory source, call `SetFinish` before constructing a CC-map iterator;
  - retain exact `DecrReadIntent` cleanup for a real in-batch continuation.
- Document that `FetchBucketData` fills a separate KV scan cache rather than the CC map.
- Keep local `docs/superpowers/` working notes ignored and outside the PR.

`HashParitionCcScanner::Merge` already provides the required frontier invariant. If it trims memory tuples behind the new pause key, it resets `memory_scan_is_finished_` to false. If the flag remains true, the snapshot-like memory source has no unconsumed tuples for the next batch, so rescanning it is unnecessary.

This reuses the existing progress flag and `SetFinish` fetch-wait/merge path. It does not add generation tracking and does not reset finished memory progress on every request reset. Table read-lock lifetime already prevents concurrent drop/recreate while the scan accesses the table.

The local and remote paths use their existing equivalent flags: `BucketScanProgress::memory_scan_is_finished_` and `RemoteScanNextBatch::memory_is_drained_`.

## Test plan

- [x] Unit/CTest coverage
- [x] Formatting/build checks
- [ ] Recovery, compatibility, or performance validation, when relevant
- [x] Documentation updated, when behavior changed

Commands and results:

```text
env LD_LIBRARY_PATH=/data/workspace/eloqkv/data_substrate/third_party/install/lib \
  cmake --build bld --parallel 4
# passed

env LD_LIBRARY_PATH=/data/workspace/eloqkv/data_substrate/third_party/install/lib \
  ./bld/tx_service/tests/TxConsistency-Test [tx]
# 121 assertions in 1 test case passed

env LD_LIBRARY_PATH=/data/workspace/eloqkv/data_substrate/third_party/install/lib \
  ctest --test-dir bld --output-on-failure --parallel 4
# 51/51 tests passed in 27.31 s

/home/ubuntu/.local/bin/clang-format --dry-run --Werror \
  tx_service/include/cc/template_cc_map.h \
  tx_service/tests/TxConsistency-Test.cpp
git diff --check
# passed
```

TDD evidence for the finished-memory regression:

- RED on the prior implementation: the supposedly finished memory source was scanned and its cache size was 1 instead of 0.
- GREEN after the skip: the memory cache stays empty, the existing semantic intent count stays at 1, and commit releases it to 0.

Not run locally: a dedicated remote/failover runtime fixture, a real asynchronous-KV callback/re-drive fixture, recovery/WAL compatibility tests, and performance benchmarks. Parent EloqKV integration is triggered separately by advancing its submodule pointer.

## Risk assessment

- Treating a completed hash memory source as snapshot-like means later CC-map mutations are not reopened after `Merge` preserves the finished flag; this is the intended `ScanBatch` behavior.
- Keeping scan ownership until finalization can increase read-set footprint and writer blocking for long locking scans.
- The local and remote skip paths are symmetric and compiled, but the remote async fetch/re-drive path has no dedicated unit fixture.
- No WAL, storage format, network protocol, or configuration changes.

## Rollback plan

Revert this PR. No migration or configuration rollback is required.

## Reviewer guide

1. `tx_service/include/cc/template_cc_map.h`: local and remote finished-memory skip after KV fetch launch and before CC-map iteration.
2. `tx_service/include/cc/ccm_scanner.h`: existing `Merge` frontier reset that makes the skip safe.
3. `tx_service/tests/TxConsistency-Test.cpp`: real CC-map regression proving empty memory cache and unchanged intent ownership.
4. `tx_service/src/tx_execution.cpp`: transaction-final scanner ownership transfer.
5. `docs/03-concurrency-control.md` and `docs/04-transaction-execution.md`: scan-source and ownership invariants.

## Follow-up work

Add a dedicated async store-handler fixture if callback/re-drive behavior needs direct unit coverage.


tx_service/tests/TxConsistency-Test.cpp
```diff
@@ -1,6 +1,14 @@
+#include <atomic>
 #include <catch2/catch_all.hpp>
+#include <vector>
```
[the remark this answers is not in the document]
liunyl: Fixed in bb6a5bb: C++ standard-library headers now precede Catch2.

tx_service/tests/TxConsistency-Test.cpp
```diff
@@ -12,10 +20,148 @@ using namespace txservice::test;
...
+
+            for (target_idx = 0; target_idx < batch.size(); ++target_idx)
+            {
+                if (batch[target_idx].key_ == target_key)
+                {
+                    found = true;
+                    break;
+                }
+            }
+        }
```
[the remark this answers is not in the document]
liunyl: Fixed in bb6a5bb: batch is cleared whenever the target is absent before the next ScanBatchTxRequest. The focused test and 20-run stress loop pass.

tx_service/tests/TxConsistency-Test.cpp
```diff
@@ -12,10 +20,149 @@ using namespace txservice::test;
...
+        release_addr.SetCceLock(1, 1, 1, 1);
+
+        TableName pooled_name(
+            std::string("a"), TableType::Primary, node.Table().Engine());
+        TableName original_name(pooled_name);
+        read_set.AddReadForRelease(release_addr, pooled_name);
+        REQUIRE(read_set.DataReadSetSize() == 1);
+
+        pooled_name = TableName(
+            std::string("b"), TableType::Primary, node.Table().Engine());
```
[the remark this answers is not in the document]
liunyl: Fixed in bb6a5bb: the test now includes string directly and keeps the required include order.

liunyl:
Reviewed the remaining outside-diff and nitpick feedback:

- The trace-enabled KickoutDataOp symbol finding was valid and is fixed in bb6a5bb.
- Direct range-middle, trailing-only, and injected error-drain lifecycle fixtures would be useful additional coverage, but the current TestNode harness only exposes a hash primary scanner and has no deterministic range or scan-error injection path. Building those facilities would be a separate engine-test project rather than a focused fix. This PR retains the deterministic commit/abort reproduction of the original race, records the exact fixture gap in the PR follow-up, and routes the reviewed cleanup paths through the same RetainScanReadForRelease and RetainScanTrailingReads helpers.

tx_service/tests/TxConsistency-Test.cpp
```diff
@@ -1,6 +1,15 @@
+#include <atomic>
 #include <catch2/catch_all.hpp>
+#include <string>
```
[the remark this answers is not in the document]
liunyl: Fixed in 3427a03: Catch2 is now a quoted third-party include, so clang-format keeps all standard-library headers before it. The test rebuilds and passes.

tx_service/tests/TxConsistency-Test.cpp
```diff
@@ -1,6 +1,15 @@
...
+#include <vector>
```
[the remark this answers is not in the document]
liunyl: Fixed in 3427a03: Catch2 is categorized after the C++ standard-library block without fighting clang-format include sorting.

tx_service/src/tx_execution.cpp
```diff
@@ -3601,75 +3534,21 @@ void TransactionExecution::ScanClose(
...
                 LockType lk_type =
                     scanner->DeduceScanTupleLockType(last_tuple->rec_status_);
-                if (lk_type == LockType::NoLock)
+                if (lk_type == LockType::NoLock &&
+                    !last_tuple->cce_addr_.Empty())
                 {
-                    drain_batch_.emplace_back(last_tuple->cce_addr_,
-                                              last_tuple->key_ts_,
-                                              PostReadType::DecrReadIntent);
+                    RetainScanReadForRelease(last_tuple, scan_table_name);
```
liangjchen: There are three functions defined: `DrainScanner`, `RetainScanReadForRelease`, `RetainScanTrailingReads`. I don't see why there need to be three. Their semantics look overlapped. To me, a single function `DrainScanner` iterates through the remaining scan tuples, add them to the read set if needed. And `DrainScanner` is done at `ScanClose`. Right now, `DrainScanner` call sites scatter in many error-handling spaces, which in my view is completely redundant: regardless of errors, `ScanClose` is always called. 

Why `DrainScanner` needs the `retain_range_last_tuple` flag? for what? 

In this specific call site, `RetainScanReadForRelease` is called here. Shouldn't there be a single call of `RetainScanTrailingReads` at the end regardless of range and hash partitioning?

tx_service/src/tx_execution.cpp
```diff
@@ -3553,46 +3517,15 @@ void TransactionExecution::ScanClose(
...
-                rw_set_.RemoveDataReadEntry(table_name, tpl.cce_addr_);
-            if (read_cnt == 0)
-            {
-                drain_batch_.emplace_back(
-                    tpl.cce_addr_, tpl.version_ts_, PostReadType::Release);
-            }
-        }
-    }
+    const TableName &scan_table_name = scan_it->second.table_name_;
+    assert(scan_table_name == table_name);
```
liangjchen: there is already `table_name`. why define, use `scan_table_name` and assert equality?

tx_service/include/tx_operation.h
```diff
@@ -430,13 +433,18 @@ struct ScanState
 
     size_t current_plan_index_{SIZE_MAX};
     std::unique_ptr<CcScanner> scanner_;
+    // References the same long-lived name storage used by regular read-set
+    // entries. Scan-close requests are pooled, so their owning copy must not
+    // become the source of a read-set string_view.
+    TableName table_name_;
```
liangjchen: Why? is this really needed?

tx_service/src/tx_execution.cpp
```diff
@@ -5847,8 +5726,40 @@ void TransactionExecution::ReleaseCatalogWriteAll(
...
+                                                    const TableName &table_name)
+{
+    if (tuple->cce_addr_.Empty() ||
+        rw_set_.GetReadCnt(table_name, tuple->cce_addr_) != 0 ||
+        cmd_set_.FindObjectCommand(table_name, tuple->cce_addr_))
+    {
+        return;
+    }
+
+    (void) rw_set_.AddRead(tuple->cce_addr_, 0, &table_name);
```
liangjchen: Using 0 for the input timestamp does not work, I think. It violates the semantics of `AddRead`. Read the function's comment.

tx_service/src/tx_execution.cpp
liunyl: Addressed in f23e1e8. The three overlapping helpers are now one `DrainScanner`: `ScanClose` invokes it once for remaining, trailing, and range-resume tuples ([ScanClose](https://github.com/eloqdata/tx_service/blob/f23e1e8e46e58b81e10338df606adf01419f38d8/tx_service/src/tx_execution.cpp#L3495-L3513), [DrainScanner](https://github.com/eloqdata/tx_service/blob/f23e1e8e46e58b81e10338df606adf01419f38d8/tx_service/src/tx_execution.cpp#L5692-L5778)). The scattered ScanNext error-path calls were removed. The boolean was renamed to `retain_range_resume_tuple` and is true only for `SlicePosition::Middle`, because only that case pins the last CCE for the next batch. The only earlier drain points left are where waiting for ScanClose would lose state: ScanOpen failure has no registered scan alias, and hash-batch cache reset discards trailing tuples.

tx_service/src/tx_execution.cpp
liunyl: Addressed in f23e1e8. The redundant `scan_table_name` local and equality assertion are gone; `ScanClose` now uses the table name passed through `ScanCloseTxRequest` directly ([lines 3495-3513](https://github.com/eloqdata/tx_service/blob/f23e1e8e46e58b81e10338df606adf01419f38d8/tx_service/src/tx_execution.cpp#L3495-L3513)).

tx_service/include/tx_operation.h
liunyl: Addressed in f23e1e8. `ScanState::table_name_` and its constructor parameters were removed. Instead, `ScanCloseTxRequest` now carries a non-owning `TableName` view of the same transaction-lifetime backing used by ScanOpen/ScanNext ([constructors and Reset](https://github.com/eloqdata/tx_service/blob/f23e1e8e46e58b81e10338df606adf01419f38d8/tx_service/include/tx_request.h#L710-L770)). The ScanOpen, ScanNext, and ScanClose request contracts now state that the backing storage must remain valid until transaction end ([ScanOpen](https://github.com/eloqdata/tx_service/blob/f23e1e8e46e58b81e10338df606adf01419f38d8/tx_service/include/tx_request.h#L458-L464), [ScanNext](https://github.com/eloqdata/tx_service/blob/f23e1e8e46e58b81e10338df606adf01419f38d8/tx_service/include/tx_request.h#L653-L659), [ScanClose](https://github.com/eloqdata/tx_service/blob/f23e1e8e46e58b81e10338df606adf01419f38d8/tx_service/include/tx_request.h#L710-L717)).

tx_service/src/tx_execution.cpp
liunyl: I checked this against the final PostRead semantics and kept 0 intentionally. These entries represent scanner-only ownership (trailing tuples or the range-resume pin), not a semantic key read, so they must be released at finalization without OCC version validation. The `PostRead` contract explicitly defines `key_ts == 0` as “does not check version stability” ([cc_handler.h lines 185-187](https://github.com/eloqdata/tx_service/blob/f23e1e8e46e58b81e10338df606adf01419f38d8/tx_service/include/cc/cc_handler.h#L185-L187)). The implementation validates only when `key_ts > 0` ([template_cc_map.h lines 1311-1312](https://github.com/eloqdata/tx_service/blob/f23e1e8e46e58b81e10338df606adf01419f38d8/tx_service/include/cc/template_cc_map.h#L1311-L1312)) and still executes `ReleaseCceLock` for `PostReadType::Release` ([lines 1382-1389](https://github.com/eloqdata/tx_service/blob/f23e1e8e46e58b81e10338df606adf01419f38d8/tx_service/include/cc/template_cc_map.h#L1382-L1389)). Passing the tuple version would incorrectly turn an implementation-only scanner pin into an additional OCC read validation.

tx_service/src/tx_execution.cpp
```diff
@@ -5885,26 +5740,41 @@ void TransactionExecution::DrainScanner(CcScanner *scanner,
...
         }
         scanner->MoveNext();
         cc_scan_tuple = scanner->Current();
     }
+
+    // Trailing tuples are scanned beyond the end key and are absent from the
+    // semantic read set. Retain their ownership before closing or reusing the
+    // scanner cache.
+    std::vector<const ScanTuple *> trailing_tuples;
+    scanner->MemoryShardCacheTrailingTuples(&trailing_tuples);
```
liunyl, line 5770: why do we still need to add trailing tuple again afte we have already drained scanner?

tx_service/src/tx_execution.cpp
```diff
@@ -5885,26 +5740,41 @@ void TransactionExecution::DrainScanner(CcScanner *scanner,
...
+        {
+            retain_for_release(tuple);
+        }
+    }
+
+    if (retain_range_resume_tuple)
+    {
+        std::vector<const ScanTuple *> last_tuples;
+        scanner->MemoryShardCacheLastTuples(&last_tuples);
+        for (const ScanTuple *tuple : last_tuples)
```
liunyl, line 5784: does hash partition scan also put a read intent lock on the last tuple for resume? If so we need to handle that too. The old code commented the resume tuple handling logic for some reason but I'm not sure if that's right

tx_service/src/tx_execution.cpp
liunyl: These are disjoint sets. `DrainScanner`'s `Current()`/`MoveNext()` iteration only covers a cache's logical `[0, size_)` range. `Merge()` can shrink `size_` with `RemoveLast()`, and the removed tail is preserved separately as `[size_, size_ + trailing_cnt_)`; `MemoryShardCacheTrailingTuples()` is the path that enumerates that tail. So this is not adding the same tuples twice. Removing the explicit trailing pass would leave ownership acquired on tuples scanned past the merge/end boundary untracked. I will make this distinction explicit in the comments and remove the stale `SetDrainCacheMode` comment.

tx_service/src/tx_execution.cpp
liunyl: Hash scans do acquire read intents, but they are not last-tuple pins that survive a completed `ScanNextBatchCc`. The current hash path uses intents only across an internal block/self-reenqueue and releases the `prior_cce`/`end_it` intents on re-entry. If the cache is full, it finishes without acquiring a new resume intent; the next transaction-level batch resumes from `BucketScanProgress::pause_key_`, not a CCE address.

Range scans are different: `ScanSliceCc` explicitly acquires a read intent on `last_cce` for `SlicePosition::Middle`, stores its lock address in the last tuple, and the next request passes that address as `prior_cce_lock`. Therefore only the range path needs the special `ScanClose` retention. The old hash block was commented out by the hash-partition scan refactor (#149), when cross-batch resume changed to the pause key. I will qualify the code/doc wording; no hash `ScanClose` branch is needed.

tx_service/src/tx_execution.cpp
liangjchen: I know what you are talking about.

tx_service/src/tx_execution.cpp
liangjchen: There is a separate issue that needs to be resolved: evaluating scan next of hash-partitioned cc map may leave orphaned locks on the cc map side.

liunyl: You were right; my previous reply conflated client-visible batch resume with resume inside one `ScanNextBatchCc`.

A hash scan does **not** retain the last returned tuple as a cross-batch resume anchor—the next client batch starts from `pause_key_`. However, when one memory pass reaches the 128-entry budget, the same CC request self-reenqueues and holds counted ReadIntents on the next unprocessed CCE and any finite-end CCE in `blocking_info_`.

That internal pin could be orphaned. The deterministic interleaving is: stale `memory_scan_is_finished_ == true` with KV still unfinished; the memory pass yields and pins; the final KV callback runs ahead of the queued continuation and marks the shard drained; the continuation then returns through `ShardIsDrained() -> SetFinish()` before the old normal-release block. This is a serialized same-shard queue-order race, not a C++ data race.

The updated branch fixes the lifecycle rather than adding a hash `ScanClose` last-tuple branch:

- `SetFinish`/`SetError` consume pending continuation/end ownership exactly once for local and remote scans.
- Normal and terminal paths use exact `DecrReadIntent`, so another intent count owned by the transaction is preserved; blocked paths release the actually granted lock type.
- `Reset` clears stale memory-finished/error state unless memory and all KV buckets are truly complete.
- Each saved wrapper address carries its generation. A transient term error can still release a matching live pin, while teardown/reuse cannot consume ownership belonging to the new object at the same address.
- Scan-error scanner ownership is retained in the transaction read set for final commit/abort cleanup.

TDD: the original reproducer was RED at 203/205 assertions; an error-code-only term fix was RED at 402/403; disabling the normal-resume generation guard was RED at 406/408. The final focused test is GREEN at 408/408, repeated 10/10, and the full suite passes 51/51. The branch and PR description are now updated at `b2d794e`.

liunyl: Corrected in `f48bff6`.

The orphan risk is real, but my previous diagnosis and the generation-based fix were unnecessarily broad. In particular, `memory_scan_is_finished_` is cross-batch scan progress, not stale request state that should be reset. The scan may enter the CC map again while KV backfill is still incomplete, and a continuation may then be pending when the shard becomes drained. This does not require recreating the table, and it is not limited to inserting 128 new keys: the 128-entry visit budget is one way to self-reenqueue, while lock/future blocking can also create a pending continuation.

The actual ownership bug is the order inside `Execute`: `ShardIsDrained()`/`SetFinish()` ran before the existing code that consumes `blocking_info_`. Therefore a resumed request could finish without decrementing its continuation and finite-end `ReadIntent` references.

The final fix is intentionally small:

- keep the term check first, so a term mismatch never dereferences old lock addresses;
- when a continuation address exists, run the existing resume cleanup before the terminal checks;
- use `DecrReadIntent` to consume exactly this request's reference, preserving any semantic intent held by the transaction;
- make the same ordering change in the local and remote paths;
- leave `memory_scan_is_finished_` unchanged and remove all generation/reset machinery from the final diff.

TDD on the original ordering was RED: the terminal path left the intent count at 2 instead of 1 (117/118 assertions). The fixed path is GREEN at 121/121, and the full suite passes 51/51.

liunyl: Corrected again in `d8b1493` after tracing `Merge`, `SetFinish`, and both local/remote hash-scan paths end to end.

My previous callback/continuation race diagnosis was wrong. A live `NoBlocking` continuation is created only when the current memory pass is unfinished, while `memory_scan_is_finished_` becomes true only when that pass reaches the CC-map end and immediately calls `SetFinish`. Therefore the state used by the earlier reproducer — finished memory plus a live continuation pin — is not produced by the real execution path.

The actual bug is simpler: on a later client batch, if memory was already finished but any KV bucket was still unfinished, `ShardIsDrained()` returned false and `Execute` entered the CC map again. That redundant memory scan could materialize the same CCE and add another `ReadIntent`.

The minimal fix is now:

- start unfinished KV fetches as before;
- if this is a first-entry path and the core memory source is already finished, call `SetFinish` before constructing the CC-map iterator;
- make the same change in local and remote paths;
- retain existing transaction-owned semantic intents until commit/abort;
- keep exact `DecrReadIntent` cleanup for genuine in-batch continuations;
- add no generation tracking and do not reset completed memory progress.

This is safe because `HashParitionCcScanner::Merge` already resets `memory_scan_is_finished_` to false whenever trimming memory tuples leaves work behind the new pause key. If the flag remains true, there are no unconsumed memory tuples for this snapshot-like scan.

TDD: the real TestNode regression was RED on the prior code because the finished memory cache contained 1 tuple instead of 0. It is GREEN after the fix: memory cache 0, existing semantic intent count remains 1, and commit releases it to 0. The focused test passes 121/121 assertions and the full CTest suite passes 51/51.

thweetkomputer (approved):

PR #531, opened 2026-07-20
Add checkpoint-stall warning and same-node-group orphan lock recovery

liangjchen:
Production incident: a leaked SCAN read intent was adopted into lock_holding_txs_ via tx-ident wrap, permanently pinning the checkpoint watermark. Checkpoints silently stopped advancing for ~5.5h until the node exhausted memory and fds. Two remediations:

1. Checkpoint-stall warning (checkpointer.cpp): CkptTsCc now reports the pinning txn/core/wlock_ts per node group. After ckpt_stall_warn_rounds consecutive rounds without watermark advance, the checkpointer logs a rate-limited WARNING naming the pinning txn, its core, and the wlock age, so a frozen pin is visible long before the OOM death spiral.

2. Orphan lock recovery for same-node-group txs (cc_shard.cpp): CheckRecoverTx now repairs registry/lock discrepancies in place (removes cce refs whose lock no longer records the txn) and, for same-ng owners, launches a pooled RecoverDeadTxCc that probes the owner shard's tx entry without blocking (Probe hop -> Resolve hop). Verdicts: Alive = leave alone; Dead (slot reused/Aborted/Finished) = release reservation-class locks (reads, read intents, data write intents) and, under skip_wal, abort-clear data write locks; Committed = release reservation-class locks, leave write locks to the owner's post-processing, bounded by committed_write_recover_window_seconds. Write locks on meta tables or with WAL enabled are handed to the log-based recovery pipeline (Sharder::RecoverTx), unchanged for remote-ng owners.

New runtime-mutable flags: ckpt_stall_warn_rounds (default 3), committed_write_recover_window_seconds (default 300).

TxConsistency-Test scenario 8 plants a dead read intent, dead write-lock bookkeeping, and a live intent on one cce, then verifies the pin is reported, dead entries are recovered through the natural checkpoint path, and the live tx survives until commit.

## Test plan

- [ ] Unit/CTest coverage
- [ ] Formatting/build checks
- [ ] Recovery, compatibility, or performance validation, when relevant
- [ ] Documentation updated, when behavior changed

Commands and results:

```text
<!-- Paste exact commands and outcomes. State checks that were not run. -->
```

## Summary by CodeRabbit

* **New Features**
  * Added enhanced checkpoint-stall diagnostics, including identifying the transaction and lock timing that pin checkpoint progress.
  * Added automatic repair and orphan-lock recovery, with staged probing and resolution for dead or inconsistent locks.
  * Introduced configurable controls for checkpoint-stall warning frequency and committed-lock recovery behavior.
* **Bug Fixes**
  * Prevented stale lock registry discrepancies from silently pinning checkpoints after ownership has ended.
  * Improved cleanup of leaked/inconsistent lock states while ensuring active transaction locks remain intact.
* **Documentation**
  * Clarified required invariants and recovery mechanisms for lock/registry correctness.
* **Tests**
  * Added coverage for orphan-lock recovery scenarios on test nodes.


tx_service/src/cc/cc_shard.cpp
```diff
@@ -60,7 +61,27 @@ DECLARE_bool(cmd_read_catalog);
...
+
+static bool ValidateRecoverWindow(const char *, int32_t)
+{
+    return true;
+}
+BRPC_VALIDATE_GFLAG(committed_write_recover_window_seconds,
+                    ValidateRecoverWindow);
+
 DECLARE_double(ckpt_buffer_ratio);
+extern bool txservice_skip_wal;
```
thweetkomputer: txservice_skip_wal is declared twice, last time is at line 64.

tx_service/include/cc/cc_shard.h
```diff
@@ -612,14 +616,64 @@ class CcShard
...
+     * whose lock no longer lists the tx are removed as in CheckRecoverTx.
+     */
+    void RecoverDeadTxLocks(TxNumber txn,
+                            NodeGroupId cc_ng_id,
+                            int64_t cc_ng_term,
+                            int64_t tx_coord_term,
+                            uint64_t wlock_ts,
+                            RecoverDeadTxCc::Verdict verdict);
+
+    uint64_t ActiveTxMinTs(NodeGroupId cc_ng_id,
```
thweetkomputer, line 674: How about removing the default value of pinning_txn and pinning_wlock_ts since both of them are set currently?

thweetkomputer (approved):

PR #553, opened 2026-08-12
Wait for metadata post-processing when closing a metadata-only transaction

liunyl:
Follow-up to #551, which merged before these two commits landed on its branch.

## 1. `CommitTx` must not fire-and-forget a metadata-only transaction

`CommitTx`'s fast path discards the caller's `CommitTxRequest` and fires the txm-owned `commit_tx_req_` without waiting, returning `true` unconditionally. For a transaction whose only state is a catalog (metadata) read that is wrong on three counts:

* an abort reports success — the fast path returns `true` even when `to_commit_` is false;
* the caller's request is never finished, so `CommitTx()`'s wrapper reads an `ErrorCode` that was never set;
* the catalog read locks are still held when `CommitTx` returns, because releasing them is part of the asynchronous post-processing the caller no longer waits for.

The third is the one that bites. A command that closes and immediately reopens a transaction — `writeConflictRetry` doing exactly that, with no backoff at all for the first four attempts — can keep a catalog entry's reader count permanently non-zero, so a DDL write intent never upgrades and the retry loop cannot converge.

Excluding `MetaDataReadSetSize()` from the fast-path guard routes those transactions through the waiting path, so the locks are released before `CommitTx` returns. The guard's other terms already cover the remaining state: `ForwardWriteCnt() > 0` implies `WriteSetSize() > 0` (both increment sites iterate the write set), and `cmd_set_` is the object-command path that only EloqKV drives.

Measured on EloqDoc's jstest suite (Debug, rocksdb, single node), before → after:

| | before | after |
|---|---|---|
| write conflicts across the run | 63,955 | **0** |
| worst single `createIndexes` | 4,755 retries | — |
| jstests step | 120 min (timed out) | completes |
| 835-test common subset | — | 2.2x faster than `main` |

`TestNodeSmoke` gains coverage for both close modes of a metadata-only transaction.

## 2. Corrected rationale for the catalog fail-fast (docs only)

The read-local comment and `03-concurrency-control` said the catalog entry is "where DDL and DML meet". It is not: every caller that asks for a catalog read-for-write is a DDL path (create/drop collection, create/drop indexes, rename), while DML and queries read the catalog without the flag. The cycle the fail-fast breaks is DDL against DDL on the same table. Behaviour is unchanged; only the description was wrong.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed transactions containing only metadata reads being incorrectly treated as empty during commit.
  * Improved handling of catalog lock conflicts so operations fail promptly instead of entering unnecessary wait states.
  * Ensured close requests complete successfully and return the expected result for metadata-only transactions.

* **Documentation**
  * Clarified catalog locking behavior, lock escalation, and retry handling for schema changes.

### DIFF
--- a/docs/03-concurrency-control.md
+++ b/docs/03-concurrency-control.md
@@ -262,15 +262,17 @@ Note that the protocol changes only the *status* of a failed acquisition, never 
 That is what lets a catalog read-for-write run under `OCC` while still taking a `WriteIntent`
 (see the read-local rule in [04](04-transaction-execution.md#isolation--protocols)).
 
-**Why catalog write intents must not wait.** A catalog entry is the one entry where DDL and DML
-meet, and an API layer typically holds its `ReadLock` on that entry before escalating to a
-`WriteIntent` on the same entry. If the escalation blocks, the resulting wait edge closes a
-cycle: the DDL holding the `WriteIntent` waits in `acquire_all_lock_op_` for `NoReadLockConflict`
-so it can upgrade to a `WriteLock`, which cannot happen until the blocked waiter releases the
-read lock it is still holding. The deadlock detector sees the cycle, but a DDL past its prepare
-log cannot be aborted, so its recovery action is a no-op and the cycle re-forms. Failing fast
-instead means the waiter never enters `blocking_queue_`, so no edge is produced and the cycle
-cannot form.
+**Why catalog write intents must not wait.** A catalog read-for-write is requested only by DDL
+paths -- an API layer's create/drop collection, create/drop indexes and rename all ask for it,
+while DML and queries read the catalog without the flag. A DDL command typically holds a lock on
+the catalog entry from an earlier step before escalating to a `WriteIntent` on the same entry, so
+two commands touching one table can each hold what the other needs. If the escalation blocks, the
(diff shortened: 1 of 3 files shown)


liunyl:
> `rw_set_.MetaDataReadSetSize()` does not cover catalog locks stored in `locked_db_`.

Confirmed and fixed in c18dd39 — the guard now also requires every `locked_db_` slot to be empty, with a comment naming both places a catalog read lock can live (`meta_data_rset_`, released by `ReleaseMetaDataReadLock()` during post-processing; `locked_db_`, released by `ReleaseCatalogsRead()` from `Reset()`) and why returning while either is held matters.

Two notes on the blast radius, since the change moves work off the immediate path:

* `locked_db_` is populated only by `Process(ObjectCommandOp&)`, i.e. the EloqKV object-command path; EloqDoc never reaches it. `FLAGS_cmd_read_catalog` does default to `true`, so it is live there.
* EloqKV's auto-commit commands do not reach `CommitTx` at all: the `already_committed` branch of `PostProcess(ObjectCommandOp&)` calls `Reset()` directly, which releases those locks synchronously. So the additional wait applies to explicit multi-command transactions, not to the per-command hot path.

PR #555, opened 2026-08-17
Ckpt ts per partition

liangjchen:
## Background

This change comes out of an extended benchmarking effort on EloqKV: driving read-only and 1:1 read/write workloads over 10M keys against a memory-limited node, measuring where throughput and tail latency actually go, and turning what we learned into targeted optimizations. Over that effort throughput improved several-fold and the tail latency that motivated the work fell from seconds to milliseconds. This PR is the tx_service half of that work; companion PRs land in brpc and eloqstore.

## The optimizations

**1. Release flush memory quota progressively instead of in one step.** A data sync round held its entire flush quota until the whole round finished. Under a large flush the shard therefore sat on memory it no longer needed while blocking admission of new writes — one of the direct causes of the multi-second p99.99 stalls this effort started from. Quota is now released per partition, weighted by bytes actually written, so admission recovers as the flush drains rather than only at the end.

**2. Publish checkpoint timestamps per partition — but only where that is durable.** Publishing early is what lets a round release its hold incrementally, and it is also how a stalled partition stops holding back the rest. The catch is that a checkpoint ts is a promise the WAL can be truncated, so it must not lead durability. `DeferCkptTsUpdate(need_persist_kv, enable_mvcc)` decides per backend:

- EloqStore, non-MVCC: publish per partition, once that partition's  `BatchWriteRecords` is durable.
- RocksDB / RocksDB-Cloud: defer to end of round — durability is established by  `PersistKV()`, not by the batch write.
- MVCC on any backend: defer — a version is durable only once base *and* archive  writes have landed.

The deferred path collects what it skipped and applies it after `PersistKV()`/`PutArchivesAll()` succeed, so both paths converge on the same state.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved flush reliability by ensuring data is durable before checkpoint timestamps are published.
  - Prevented stale-term data from affecting checkpoint updates.
  - Preserved accurate error reporting when storage writes or flushes fail.
  - Improved cleanup and recovery after flush completion or failure.

- **Performance**
  - Added progressive release of flush resources as partitions become durable.
  - Improved partition-level flush progress reporting.

- **Documentation**
  - Updated durability, recovery, and checkpoint-flush documentation.


store_handler/rocksdb_handler.cpp
```diff
@@ -516,9 +516,13 @@ bool RocksDBHandler::PutAll(
         &batch,
     const std::function<void()> *yield_fptr,
     const std::function<void()> *resume_fptr,
-    const std::function<void()> *sync_yield_fptr)
+    const std::function<void()> *sync_yield_fptr,
+    const std::function<void(uint64_t, uint64_t)> *partition_progress_fptr)
```
thweetkomputer, line 520: This signature now participates in the new per-backend checkpoint publication contract, but the implementation below still writes every `flush_task_entry`. `DataStoreServiceClient::PutAllImpl` filters merged flush entries with `FindNewestTerms` / `IsNewestTerm` before datastore IO; RocksDB should do the same in both loops. Since this backend returns `NeedPersistKV() == true`, deferred ckpt-ts publication can skip stale entries while those stale records have already been written and later persisted by `PersistKV()`.

liangjchen: The rocksdb handler is obsolete and should only be used as a standalone, testing environment (in which case there is no term or failover). All data stores are expected to be hooked via Data Store Service (DSS), which already supports rocksdb, rocksdb-cloud and eloqstore.  So, per-partition checkpoint ts report is expected to be in the DSS path only.

thweetkomputer:
Conflict with main branch, please resolve.

store_handler/rocksdb_handler.cpp
thweetkomputer: Sure, we need to migrate the standby from pure rocksdb handler to DSS rocksdb later.

liangjchen:
> Conflict with main branch, please resolve.

Rebased w/ the main. The new commit #556 fixes a bug that fails to update dirty key status ``BeingCkpt``, causing range repartitioning and bucket migration to hang indefinitely. The fix is a little complex and built on an overly-complex prior design. The rebase uses a much simpler fix.

tx_service/src/cc/local_cc_shards.cpp
```diff
@@ -6053,10 +6053,73 @@ void LocalCcShards::FlushDataImpl(FlushDataTask *cur_work,
...
+            return;
+        }
+        // 128-bit intermediate: quota and byte totals are both full-width
+        // uint64 counters, so the product can exceed 64 bits.
+        const uint64_t target = static_cast<uint64_t>(
+            static_cast<unsigned __int128>(quota_progress.task_flush_quota_) *
+            done_bytes / total_bytes);
+        if (target > quota_progress.released_)
+        {
+            quota_progress.mem_controller_.DeallocateFlushMemQuota(
```
thweetkomputer, line 6094: The quota should not be returned until the corresponding memory is actually released. This callback only decrements `DataSyncMemoryController` accounting; `cur_work->flush_task_entries_` still owns all of its `data_sync_vec_` / `FlushRecord` buffers, and this function reads them again below before the task is destroyed. Since `DeallocateFlushMemQuota()` wakes data-sync scanners, a slow sibling partition can admit new scan buffers while the old task memory is still resident, allowing actual checkpoint memory to exceed the configured `ckpt_buffer_ratio`. The failure path compounds this by reporting a partition’s full `serialized_bytes_` even when later batches were never sent. Please either free/detach the completed partition’s buffers and account their actual in-memory bytes before reporting progress, or retain the quota until `PutAll`/the flush task releases those buffers.

thweetkomputer (approved):

PR #569, opened 2026-09-09
fix: release replaced standby command storage

thweetkomputer:
## Context

A standby forward entry can accumulate large serialized commands before an internally generated overwrite replaces them. Protobuf `clear_cmd_list()` removed the logical commands but retained their strings and capacities in that live entry. These obsolete bytes were also absent from the history buffer's `ByteSizeLong()` accounting.

## Behavior before and after

Replacing eight 1 MiB commands with a small overwrite now releases the replaced command storage immediately. Routing/version fields and the overwrite-plus-following-command sequence stay unchanged; the wire format is unchanged.

## Implementation

Swap the old repeated string field with an empty non-arena field in `StandbyForwardEntry::AddOverWriteCommand`, then serialize the overwrite. The temporary field destroys the replaced strings before a new payload is added. Keep the protobuf storage-retention rationale in the source comment beside the swap.

Release only the obsolete command field. Rebuilding the whole message would unnecessarily replace routing/version metadata. This runs on the owner shard while constructing the forward entry, before it is handed to the stream/history path. The change does not redefine total history memory accounting or allocator RSS behavior.

## Test plan

```text
git diff --exit-code d1acea36aac64bd92ea14541096bed4e55d8fe24 HEAD -- . ':(exclude)docs/**'
```

The prior PR record reports 30 assertions in 4 cases and CTest 4/4 passing on commit `c320f94a99fc3717a9ae3e69d37c0f5aad407cb7` (base `ee0f476`). Its baseline control retained 8,389,192 bytes against a 4,344-byte limit and failed as expected.

## Risk assessment

The replaced field is exclusively owned by the forward entry; no new concurrency or durability boundary is introduced. Large-to-large overwrites may allocate a fresh command string instead of reusing an obsolete capacity. The result's live capacity and other protocol bookkeeping may still exceed serialized size.

## Rollback plan

Revert this commit. No configuration, protocol, or persisted-data migration is required.

## Reviewer guide

Review `StandbyForwardEntry::AddOverWriteCommand` and the two new cases in `StandbyForward-Test.cpp`: one measures actual protobuf-owned storage after repeated replacement, the other round-trips the message and verifies routing metadata and command order.

## Follow-up work

Candidate cancellation/history GC and terminal replay ownership are separate changes; this PR does not address them.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved standby replication handling when overwrite commands replace earlier commands, preventing unnecessary retained memory.
  - Preserved routing and version information during overwrite processing.
  - Ensured commands added after an overwrite continue to be retained correctly.
  - Corrected remote forwarding so it carries the executed command state rather than an outdated pre-execution state.

### DIFF
--- a/docs/standby_replication_protocol.md
+++ b/docs/standby_replication_protocol.md
@@ -31,7 +31,9 @@ Committed object changes are captured on the owner shard's apply path. A
 forwarded command represents the command after owner-side execution, because
 the standby performs commit replay and does not rerun the command's execution
 phase. Whole-object replacement is available for changes that cannot safely be
-expressed as an incremental command.
+expressed as an incremental command. It releases the superseded commands while
+preserving the entry's routing and version metadata. Subsequent commands follow
+the replacement in replay order.
 
 Each `CcShard` is an independent sequence group and assigns monotonically
 increasing forward sequence ids. This provides FIFO ordering within a shard and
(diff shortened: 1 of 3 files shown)


docs/standby_replication_protocol.md
```diff
@@ -31,7 +31,9 @@ Committed object changes are captured on the owner shard's apply path. A
 forwarded command represents the command after owner-side execution, because
 the standby performs commit replay and does not rerun the command's execution
 phase. Whole-object replacement is available for changes that cannot safely be
-expressed as an incremental command.
+expressed as an incremental command. It releases the superseded commands while
+preserving the entry's routing and version metadata. Subsequent commands follow
+the replacement in replay order.
```
liunyl: no need to put the detailed change in architecture docs

liunyl (approved):

PR #570, opened 2026-09-09
fix: reclaim standby history after candidate removal

thweetkomputer:
## Context

A candidate standby can be the only consumer protecting large forward entries in a primary shard's history. Cancelling that candidate, or removing it after memory-cap eviction, previously erased the candidate but left otherwise-unneeded entries owned by the history queue. Successful writes do not schedule the retry cleanup, so the retained entries could remain idle until another unrelated cleanup trigger.

## Behavior before and after

Cancellation and cap-driven candidate removal now reclaim history that no remaining candidate or subscribed standby needs. Promotion preserves bootstrap messages by installing the subscriber watermark before removing its candidate on the same shard. Subscriber term acceptance, wire format, sequence allocation, and retry scheduling remain unchanged.

## Implementation

- Run existing watermark GC after a candidate is removed explicitly.
- Use `PromoteCandidateStandby` from `ResetStandbySequenceId` to preserve the required ownership order.
- Run watermark GC once after cap eviction removes affected candidates.
- Explain cleanup triggers and promotion ordering beside the affected shard code and API.

GC reuses the existing minimum-needed-sequence rule across both subscribed and candidate consumers. Merely adding GC to `RemoveCandidateStandby` with the original RPC order would temporarily leave a promoted node unprotected and discard its bootstrap history, so promotion is a single owner-shard API. The cap path finishes its candidate-map iteration before invoking GC and never reuses `entry_raw` afterward.

## Test plan

```text
git diff --exit-code 158a2e6f82e99fdce7268bb7c716be7b08f1cb90 HEAD -- . ':(exclude)docs/**'
```

The prior PR record reports 118 assertions in 5 history cases and CTest 7/7 passing on commit `e2c686fdfed7bfb07b12db69bf8e40a98df3ec9a` (base `ee0f476`). The baseline cancellation and cap-removal controls retained sequences {1, 2, 3} and {2, 3}, respectively, and failed their release assertions as expected.

## Risk assessment

Cleanup can perform more destructor work in the candidate-removal callback, bounded by the already-retained history. All operations occur on the owner shard. The subscriber must be installed before candidate GC during promotion; remaining consumers must continue protecting their requested sequence ranges. This change does not alter WAL/checkpoint durability, fencing rules, or the serialized-byte history quota, and does not claim to diagnose any observed production timeout.

## Rollback plan

Revert this commit. No data/configuration migration is required.

## Reviewer guide

Review `RemoveCandidateStandby`, `PromoteCandidateStandby`, and the cap-eviction tail in `ForwardStandbyMessage`; then the `ResetStandbySequenceId` call site and the five real-queue regressions. The test peer exposes queue/index/accounting state only to the test.

## Follow-up work

Overwrite protobuf capacity and replay-terminal ownership are separate changes. Initial-subscription retry scheduling and broader generation fencing are unchanged.

### DIFF
--- a/docs/standby_replication_protocol.md
+++ b/docs/standby_replication_protocol.md
@@ -47,13 +47,8 @@ primary retains messages needed by a failed send or a bootstrapping candidate.
 
 The primary-side history is a bounded recovery backlog, not an unbounded audit
 log. Entries that every subscriber has passed and no candidate still needs are
-released. Cancelling a candidate or removing it after history eviction also
-releases entries that no remaining consumer needs. Promotion establishes the
-subscriber's history protection before removing its candidate on the owner
-shard, so bootstrap history remains owned throughout the transition.
-
-If a send fails, the primary retries from the subscriber's next sequence id
-while the required entries remain buffered.
+released. If a send fails, the primary retries from the subscriber's next
+sequence id while the required entries remain buffered.
 
 Memory pressure may evict an entry that a subscriber still needs. In that case
 the primary sends an explicit out-of-sync indication and stops advancing the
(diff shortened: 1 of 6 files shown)


liunyl (approved):

docs/standby_replication_protocol.md
```diff
@@ -47,8 +47,13 @@ primary retains messages needed by a failed send or a bootstrapping candidate.
 
 The primary-side history is a bounded recovery backlog, not an unbounded audit
 log. Entries that every subscriber has passed and no candidate still needs are
-released. If a send fails, the primary retries from the subscriber's next
-sequence id while the required entries remain buffered.
+released. Cancelling a candidate or removing it after history eviction also
+releases entries that no remaining consumer needs. Promotion establishes the
```
liunyl: no need to put the detailed change in architecture docs

thweetkomputer: fixed

PR #572, opened 2026-09-09
fix: refresh defragmented keys in place without leaking temporary objects

thweetkomputer:
## Context

Heap defragmentation released a temporary key's unique_ptr before moving its contents into the page. The separately allocated KeyT object then had no owner and leaked on each qualifying relocation.

## Behavior before and after

Page keys retain their values while key-owned storage is refreshed. Relocation no longer allocates a separate temporary KeyT object or erases/reinserts the key. This applies to versioned and non-versioned entries. Defrag eligibility, locking checks, and wire/storage formats are unchanged.

## Implementation

Replace clone/erase/emplace in Iterator::DefragCurrentIfNecessary with `*key = KeyT(*key)`. Copy construction completes while the original value is valid, then rvalue assignment replaces the existing slot. The expression's temporary is automatically destroyed; there is no named key_clone/key_copy or owning pointer.

Register CcPageDefrag-Test to verify balanced key lifetimes and unchanged key/record contents, including repeated first/middle/last relocations. Full-page long-string tests additionally verify refreshed payload storage, unchanged vector allocation/capacity, and a stable iterator key address.

In-place replacement avoids shifting unrelated keys. Inserting before erasing could grow a full page's vector and invalidate element addresses; the final implementation does not change vector size. Key-owned payload allocations through the copy constructor remain necessary.

The long-string fixture retains production EloqStringKey copy/move semantics and forces defrag eligibility through the real TxKey interface, independently of allocator occupancy. Local ordering rationale is documented beside the code; no architecture change is needed.

## Test plan

- [x] Focused unit coverage
- [x] Formatting/build checks
- [ ] Recovery, compatibility, or performance validation
- [x] Local ownership/ordering comments updated

Commands run on 2026-09-10 using the existing isolated aarch64 S3 Debug configuration based on engine main `5c0be9f`:

```sh
LD_LIBRARY_PATH=/opt/eloq/third_party/lib:/opt/eloq/third_party/lib64 cmake --build /tmp/eloq-defrag-key-fix.qel3z8fg/bld --target CcPageDefrag-Test CcEntry-Test --parallel 2
python3 /tmp/eloq-defrag-key-fix.qel3z8fg/review_inplace_run_checks.py
# From the source checkout:
clang-format-18 --dry-run --Werror tx_service/include/cc/template_cc_map.h tx_service/tests/CcPageDefrag-Test.cpp
git diff --check 5c0be9f09b40b3358c14b7401765127b1f4210ed
```

Build passed. CcPageDefrag-Test passed 1290 assertions in 4 cases; CcEntry-Test passed 186 assertions in 5 cases. Both baseline-template lifecycle controls failed as expected at 4 live key objects versus 3 page keys. Formatting and merge-base diff checks passed.

No fresh configure, HA, live-server workload, TCL, performance benchmark, or sanitizer run was performed in this follow-up.

## Risk assessment

Storage refresh now uses KeyT copy construction and rvalue assignment instead of vector erase/emplace. Production string-key semantics are tested with long strings on full pages for both entry variants. Lifecycle tests retain coverage of the original leak. Live defrag scheduling and allocation pressure were not exercised. No concurrency or persistence boundaries change.

## Rollback plan

Revert this PR; no data migration is required.

## Reviewer guide

Start at Iterator::DefragCurrentIfNecessary and CcPageDefrag-Test.cpp. Check construction before replacement, refreshed string storage, stable page allocation, and balanced lifetimes. The final diff is limited to the production template, focused test, and CMake registration.

## Follow-up work

None in this PR. The defect requires enabled defragmentation and is not evidence that it caused the reported production value-buffer retention.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved key relocation during page defragmentation to preserve key and record data while managing temporary copies safely.

- **Documentation**
  - Clarified ownership and cleanup behavior for temporary keys during relocation.

- **Tests**
  - Added coverage for defragmenting first, middle, and last entries.
  - Verified behavior across versioned and non-versioned pages, including data ordering and resource cleanup.

### DIFF
--- a/tx_service/include/cc/template_cc_map.h
+++ b/tx_service/include/cc/template_cc_map.h
@@ -9098,14 +9098,10 @@ protected:
             TxKey tx_key = TxKey(key);
             if (tx_key.NeedsDefrag(heap))
             {
-                // Copy before erase invalidates key; the temporary needs no
-                // separate heap allocation.
-                KeyT key_copy(*key);
-                auto &keys = current_page_->keys_;
-                auto it = keys.begin() + idx_in_page_;
-                assert(it != keys.end());
-                it = keys.erase(it);
-                it = keys.emplace(it, std::move(key_copy));
+                // Copy construction refreshes key-owned storage before the
+                // old value is released. Assign in place to preserve page
+                // positions and avoid growing the key vector on a full page.
+                *key = KeyT(*key);
                 defraged = true;
             }
 
--- a/tx_service/include/cc/template_cc_map.h
+++ b/tx_service/include/cc/template_cc_map.h
@@ -9103,7 +9103,8 @@ protected:
                 auto it = keys.begin() + idx_in_page_;
                 assert(it != keys.end());
                 it = keys.erase(it);
-                it = keys.emplace(it, std::move(*(key_clone.release())));
+                // The page takes the key contents, not the temporary object.
+                it = keys.emplace(it, std::move(*key_clone));
                 defraged = true;
             }
 
--- a/tx_service/include/cc/template_cc_map.h
+++ b/tx_service/include/cc/template_cc_map.h
@@ -9098,13 +9098,14 @@ protected:
             TxKey tx_key = TxKey(key);
             if (tx_key.NeedsDefrag(heap))
             {
-                auto key_clone = std::make_unique<KeyT>(*key);
+                // Copy before erase invalidates key; the temporary needs no
+                // separate heap allocation.
+                KeyT key_copy(*key);
                 auto &keys = current_page_->keys_;
                 auto it = keys.begin() + idx_in_page_;
                 assert(it != keys.end());
                 it = keys.erase(it);
-                // The page takes the key contents, not the temporary object.
-                it = keys.emplace(it, std::move(*key_clone));
+                it = keys.emplace(it, std::move(key_copy));
                 defraged = true;
             }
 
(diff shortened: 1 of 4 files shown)


tx_service/include/cc/template_cc_map.h
```diff
@@ -9103,7 +9103,8 @@ class TemplateCcMap : public CcMap
                 auto it = keys.begin() + idx_in_page_;
                 assert(it != keys.end());
                 it = keys.erase(it);
-                it = keys.emplace(it, std::move(*(key_clone.release())));
+                // The page takes the key contents, not the temporary object.
+                it = keys.emplace(it, std::move(*key_clone));
```
liangjchen: ``key_clone`` is not necessary. ``keys.emplace`` already allocates the key w/ copy constructor. Imagine that ``KeyT`` is ``std::string`` and each key is only 8 bytes. There is small object optimization, and ``key_clone`` is useless.

liangjchen (approved):

tx_service/include/cc/template_cc_map.h
thweetkomputer: Agreed — removed the explicit clone entirely in ed4ed35. The key path is now simply `*key = KeyT(*key)`: copy-construct the new value, then assign it into the existing slot.

This also removes erase/emplace, avoiding both the separately allocated KeyT object and vector growth or shifts when the page is full. The expression's temporary is automatically destroyed.

Added full-page long-string coverage using the production EloqStringKey copy/move semantics. It verifies that the string buffer changes while vector storage, capacity, key contents, and the iterator's key address are preserved. CcPageDefrag-Test passed 1290 assertions in 4 cases; CcEntry-Test passed 186 assertions in 5 cases. Both original-template lifecycle controls still detect the leak as expected.

PR #573, opened 2026-09-09
fix: reset pooled remote upload command images

thweetkomputer:
## Context

RemoteUploadTxCommandsCc is pooled for commands uploaded to a bucket's new owner. Reset appended into cmds_vec_ without removing the previous request's commands, so a later upload could expose old commands to ObjectCcMap and retain every prior command image.

## Behavior before and after

Each Reset now exposes exactly the current upload's command list. Completion destroys its command strings before returning the wrapper to the pool, including abort completion; an idle wrapper no longer retains its last command images. The message format is unchanged.

## Implementation

Clear cmds_vec_ before populating Reset. Override Free to clear its owned strings before the base release-store publishes the slot as idle. Add request-reuse and completion-boundary tests. Explain request isolation and cleanup-before-pool-publication ordering in nearby source and API comments.

Reset enforces independent request contents even when called without an intervening pool return; terminal Free additionally releases payloads when a slot is never reused. Keep vector slot capacity for reuse. Cleanup remains after completion callbacks and does not affect pending execution. ObjectCcMap has already decoded the serialized strings into independently owned commands before completing the upload.

## Test plan

```text
git diff --exit-code 4fe82da39749a898bdcd51ac2cb9e08da93a106a HEAD -- . ':(exclude)docs/**'
```

The prior PR record reports RemoteUploadTxCommands-Test passing 36 assertions in 2 cases and CcMessagePool-Test passing 5 assertions in 1 case on commit `b50ce50` (base `ee0f476`). Independent controls without Reset or Free cleanup failed the current-command-list and post-completion-empty assertions as expected.

## Risk assessment

Cleanup must remain after the consumer and completion callback finish and before pool publication. The change does not alter lock acquisition, commit decisions, command decoding, or transport retry ownership. Tests exercise the request/result lifecycle but do not execute a full migration across live nodes.

## Rollback plan

Revert this PR; no storage migration is required.

## Reviewer guide

Review RemoteUploadTxCommandsCc::Reset/Free and the consumer's UploadTxCommandsCc::CommandList contract first, then the reuse and abort-boundary assertions in RemoteUploadTxCommands-Test.cpp.

## Follow-up work

None in this PR. Other remote request pools and ordinary standby command forwarding are outside this change.

### DIFF
--- a/docs/08-range-and-bucket-management.md
+++ b/docs/08-range-and-bucket-management.md
@@ -81,11 +81,6 @@ pending bucket version so retries or delayed messages cannot overwrite a newer
 ownership decision. Staged migration logs let replay resume work that crossed
 a leader failure.
 
-A remote command upload owns only its current request's serialized commands.
-The receiver retains those images through consumption and the completion
-callback, then destroys them before publishing its pooled wrapper as reusable.
-Decoded commands require independent ownership if they outlive that upload.
-
 ## Secondary-index integration
 
 Building a secondary index traverses primary-key ranges because those ranges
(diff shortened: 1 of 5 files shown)


liunyl (approved):

docs/08-range-and-bucket-management.md
```diff
@@ -81,6 +81,11 @@ pending bucket version so retries or delayed messages cannot overwrite a newer
 ownership decision. Staged migration logs let replay resume work that crossed
 a leader failure.
 
+A remote command upload owns only its current request's serialized commands.
```
liunyl: no need to put the detailed change in architecture docs

